PR #18, opened 2017-09-30
Add -expert build option

akhuettel:
Hi. Gentoo klayout maintainer here... It would be very useful for us to be able to freely configure CFLAGS, LDFLAGS, ... (actually our QA guidelines more or less require that), and this is what the attached commits allow. Adding it "upstream" makes much more sense to me than patching every new klayout version on our side, and maybe also somebody else will find it useful. 

The first commit improves build.sh so it correctly handles qmake arguments containing spaces. Unfortunately I haven't found any way to do this correctly without relying on bash-specific features. The logic mirrors what we have in Gentoo in our qt-package build scripts.

The second commit adds an option -expert. It overrides completely any flags set by -debug or -release, and sets the build parameters from autotools-standard variables, CC, CXX, CFLAGS, CXXFLAGS and so on. Again the logic mirrors precisely what the Gentoo qt-package build scripts do. In addition it disables stripping of the binaries, which can be done later separately if needed.

Testing done: Built succuessfully with 
CFLAGS="-march=native -O2 -pipe -g"
CXXFLAGS="-march=native -O2 -pipe -g"
LDFLAGS="-Wl,--as-needed -Wl,-O1 -Wl,--hash-style=gnu"
See the following link for an example how this ties into our build system:
https://gitweb.gentoo.org/repo/gentoo.git/tree/sci-electronics/klayout/klayout-9999.ebuild?id=770e5d6fdf74b2312b2049fe471c7c9c031486b4


klayoutmatthias:
Thanks for bringing this up. I am aware that the defaults are inherited by qmake. So far, this wasn't an issue, but it's surely good to have a way to override the compile flags. 

I wonder why -qt5, -qtlib etc. are still required. I thought they should be taken all from qmake. At least -qt5 should be auto-detected now.

akhuettel:
> I wonder why -qt5, -qtlib etc. are still required. I thought they should be taken all from qmake. At least -qt5 should be auto-detected now.

You mean in our Gentoo ebuild? They probably are. I only just modified the script from the previous (0.24, qt4-only) version, and may not have fully completed the cleanup yet. Right now I'm very happy it builds fine against qt5 (since we get pressure from our qt team, which wants to drop qt4 from the distribution). More cleanups to come; compared to the version in the link I have also in the meantime enabled linking to python.

PR #31, opened 2017-12-06
Fix compilation with `-with-64bit-coord`

pankajp:
`src/db/unit_tests` still do not compile so need to fix it or temporarily remove it
from `src/db/db.pro` to make klayout compile with the experimental 64bit coordinate support.


klayoutmatthias:
Thanks for taking care of this.

PR #56, opened 2018-01-03
Using non-OS-standard script languages

Kazzz-S:
Added Ruby2.4 and Python3.6 from MacPorts for testing.

Still have issue mentioned in Kazzz-S#19


Kazzz-S:
Withdraw this PR.

Issue #85, opened 2018-03-01
disable python "Caught the following exception:" window in Macro Development


lukasc-ubc:
Hi @klayoutmatthias,

This question is regarding handling try/except cases in python modules loaded in KLayout.

On MacOS (I haven't checked on Windows), when a module is within the KLayout python folder, e.g., urllib3 in ~/.klayout/python/urllib3, and when you run a macro that uses that module, with the Macro Development window open, each try/except creates exceptions which pop up a GUI:

> Caught the following exception:
> ....
> Press 'Ok' to continue and 'Cancel' to stop in the debugger

However, if that module is installed in the system, e.g., pip install urllib3, then the same exceptions don't trigger the GUI.  

And with the Macro Development window closed, you don't see these message and everything works fine.

What this means is that for modules that I develop and put in the KLayout path, I need to minimize the use of try/except, otherwise it makes debugging difficult (if you have to click "OK" 20 times before you get to the actual error you want to debug).

Is there any way to selectively disable the try/except dialogs for specific modules, or perhaps placing them in folders (salt/...) so they don't have these messages?

thomaslima:
Seems intentional...

Found the behavior in https://github.com/klayoutmatthias/klayout/blob/master/src/lay/lay/layMacroEditorDialog.cc#L2804-L2816

Note: it seems you can disable it in a macro debug setup dialog cf. https://github.com/klayoutmatthias/klayout/blob/master/src/lay/lay/layMacroEditorDialog.cc#L1975

lukasc-ubc:
Thanks Thomas.  

Interesting that it was designed this way, where other Python environments allow the code itself to deal with try/except cases, and don’t pop up GUIs.

It doesn’t look like the “config_set” method is exposed to Python, so perhaps we can't turn it off there?

> dir(pya.Application.instance().main_window().instance())

[...]

PR #93, opened 2018-03-15
(macOS) fixing the bug mentioned in PR #85

thomaslima:
Finally figured it out. Check the offending line. Don't know how that did not fail in linux or windows! PR #85


thomaslima:
Just tested it.

Fixed by @klayoutmatthias in 1306229. Thanks!
Mattias, I think you should hit the button merge for the sake of documenting this.

thomaslima:
@klayoutmatthias in situations like this, you could have directly pushed to my branch because I had activated the "allow edits from maintainers" option:

screen shot 2018-03-16 at 12 26 22

Thanks!

PR #96, opened 2018-03-16
(macOS) Adding support for qt and python3 installed from brew

thomaslima:
This is still not recommended for release .dmg because it *might* break if the builder's python is different than the user's python version. E.g. built on 3.6.4_4 and user has 3.7.0 from brew.

- Build process for macOS greatly simplified. Takes about one hour total.
```bash
brew install python3 qt
./build4mac.py -p B36 -q Qt5Brew
./build4mac.py -p B36 -q Qt5Brew -y
# This places a klayout.app inside qt5.pkg.macos-HighSierra-release
```
- release packages built from this can be run in any mac so long as they have installed python3 (qt not needed) from brew.
- if installing for yourself, I recommend using `-Y` flag, and changing `build4mac_env.py` variables to those set in commit fa96f6d
- from here on, we can create a brew formula with klayout for mac users. Installation would then be simplified to:
`brew install klayout` or `brew install klayout-dev`


thomaslima:
I apologize for the commit history mess. I promise to rebase the PR branch in the future.

thomaslima:
In fact let me do it right now.

thomaslima:
New pull request in #97

PR #143, opened 2018-07-16
Pymod improvements for MacOS

thomaslima:
- Used @rpath flags for dynamic libraries
- changed setup.py code to 4-spaced (sorry)
- using setuptools instead of distutils
- adding compilation job to travis-ci (https://travis-ci.org/lightwave-lab/klayout)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -128,8 +128,7 @@ class Config(object):
             args = []
             if mod[0] == "_":
                 args += ["-Wl,-dylib", '-Wl,-install_name,@rpath/%s' % self.libname_of(mod)]
-            else:
-                args += ['-Wl,-rpath,@loader_path/']
+            args += ['-Wl,-rpath,@loader_path/']
             return args
         else:
             # this makes the libraries suitable for linking with a path -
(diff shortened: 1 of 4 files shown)


atait:
Would it be possible to revert 7540da8, then reapply as two patches: whitespace and non-whitespace changes? Not that necessary, just git book keeping...

setup.py
```diff
@@ -31,122 +31,124 @@
...
+            # Avoids many "type-punned pointer" warnings
+            return ["-Wno-strict-aliasing"]
+
+    def link_args(self, mod):
+        """
+        Gets additional linker arguments
+        """
+        if platform.system() == "Windows":
+            return []
+        elif platform.system() == "Darwin":
```
atait: The changes only affect macosx. Was this already working for linux? If not, maybe we could wait to merge until linux is also working, because I think it is very close.

thomaslima:
@atait feel free to clean this PR! =)
@klayoutmatthias also has push permissions to my branch while this PR is open (magic of GitHub)

thomaslima:
All we would need for linux is to replace @loader_path with $ORIGIN. But I assumed that it already worked for linux. We also have to remove somehow setting `runtime_library_dirs`, which might not be necessary if we use @rpath strategy.

thomaslima:
Also, building without expat or curl libraries worked (https://travis-ci.org/lightwave-lab/klayout/builds/404476704). So I worry that we are not battle testing this enough.

PR #145, opened 2018-07-19
Travis CI updates

dwang:
- Ubuntu builds are added to travis.yml (Python 2.6, 2.7, 3.3, 3.4, 3.5, and 3.6)
- macOS 10.12 and macOS 10.11 are added to travis.yml
- Uses `distutils.sysconfig` instead of `sysconfig` to add support for Python 2
- Sets ext_suffix to ".so" if it is null
- Uses a different method to get the time because `clock_gettime` is not implemented in macOS 10.11 and lower

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,16 @@
-language: python
-python:
-  - "3.3"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-  - "3.6"
-  - "3.7" 
-
 matrix:
   include:
     - os: linux
       dist: trusty # Ubuntu 14.04
       sudo: false
+      language: python
+      python:
+        - "3.3"
+        - "3.4"
+        - "3.5"
+        - "3.6"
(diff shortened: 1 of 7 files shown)


thomaslima:
- Also added a missing include in tlLog.cc: stdlib.h, whose absence causes `getenv` to fail.

Thanks @dwang.

thomaslima:
I noticed that the wheel files for Mac were around 8MB, whereas those for linux were 70MB. Any ideas why? Here's where the builds from our travis-ci account are going (extract files from wheel with `tar vzxf`, or test install with `pip install wheel.whl`): https://www.dropbox.com/sh/rorqqlezkubqw7p/AACk2ThizQ3d-weKRTWxC_jXa?dl=0

thomaslima:
Passing build: https://travis-ci.org/lightwave-lab/klayout/builds/406039644

PR #168, opened 2018-09-18
Travis automation formulas for macOS builds

thomaslima:
Hi Matthias,

I have finally finished a stable automation formula for building for macOS. Here's what it does:

1. Builds klayout for macOS El Capitan(Xcode8), Sierra (Xcode8.3) and High Sierra (Xcode9.4), and a combination of System's python (2.7) and Brew's 3.7. Total of six images. All with brew's latest qt, right now 5.11.1.
2. Packages the app bundle into a compressed, read-only dmg file.
    - Here, I have simplified Kazzz's process as such. I prepared a template .dmg file using his script `makeDMG4mac.py`, but with empty klayout.app bundle and klayout.scripts folder. Then I populate this .dmg file after build using `hdiutil` directly. The image name is like this `qt5.pkg.macos-ElCapitan-B37-release-0.25.4-9a24d32`.
3. Uploads contents of folder `deploy/` to a dropbox folder `/Builds/klayout/` via a developer key. For that to work, it suffices to set up an environment variable in travis called `DROPBOX_OAUTH_BEARER`, as documented [here](https://github.com/Jawnnypoo/dropbox-deployment). There should be a log file, an md5 file and a dmg file.

To use in main klayout project:
- Setup an account in travis-ci.org and attach it to your GitHub.
- Configure the build setup. It could be at every commit, or at every day/week/month on master as a cron job, or at every pull request etc.
- If you want it deployed to your dropbox, it suffices to set the key above. Otherwise, you can add any script to upload to klayout's server. Or even GitHub releases, which offers hosting for free for built images. The later two suggestions require changing the .travis.yml file.
- If you run into maximum time issue, please ask their team to increase the build time. Right now it is taking 1h30m to build without tests. I will ask them to increase the build to my fork from 2 to 3h to fit the tests.
- I will continue building master every week on my fork and upload to the dropbox folder below.

Minor Issues/Future Work:
- The root folder is getting cluttered with scripts, Makefile, yml configuration files etc. Maybe one day it would be nice to centralize all into a Makefile for unix-like systems.
- The travis.yml file will collide with the one in pymod branch. I can fix the merge conflicts when the time arises.
- The script will always pick up the latest qt version from [home-brew's repo](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt.rb). When that changes, we have to brace for potential bugs. Alternatively, we can create our own cask so that we can phase in dependency upgrades when we're ready.
- The script will also always pick up the latest python's version (3.7). Other parts of the script are hard-coded to 3.7. When python reaches 3.8, the script will break again. We need to edit the build script to be more automatic with respect to finding python's sources.

Bonus:
- I will put all my built images in the following link. Right now there's only the 0.25.4 version: https://www.dropbox.com/sh/al4a36sryc0vc7g/AADzHicpL6EUEB2gik-Keb1La?dl=0


thomaslima:
Please do not merge yet until I've had the chance to test the builds. I welcome comments though.

thomaslima:
Ready to merge. I've tested the generated images in my Mac with High Sierra and two VMs with Sierra and El Capitan. The tests were very shallow, I opened klayout without errors, and tested whether the python executable and numpy were bundled with klayout.

thomaslima:
I feel like your last comment was cutoff. I think it applies to master, because it automatically builds klayout with qt5.11 for mac for release in the website. It's much better than doing it in my computer with uncontrolled environment.

PR #179, opened 2018-10-05
Fixed plugin detection in MacOS

thomaslima:
- The GDS/OASIS etc streamers were moved to another folder next to klayout, which required a change in the way klayout.app and ut_runner were linked.
- This patch also fixes a bug in one of the travis images with osx10.11. Travis changes specific things like default ruby version in their images at random times. This is annoying and has no immediate solution except for obviously moving away from Travis.

Most changes were done by @klayoutmatthias . Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ matrix:
       os: osx
       osx_image: xcode8  # macOS 10.11
       env:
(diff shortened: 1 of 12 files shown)


klayoutmatthias:
Hi Thomas,

the warnings may happen if one of the Qt binding lib isn't loaded. Maybe it was missing.

I'll merge anyway ... let's see what my builds say.

Thanks and best regards,

Matthias

Issue #160, opened 2018-08-29
MacOS build for 0.25.4


Kazzz-S:
@klayoutmatthias 

Thank you for making the maintenance release of 0.25.4!
I've built different binary packages for MacOS as usual and available below.
https://www.dropbox.com/sh/82bqswuidiuakks/AAAYUP6OuVWn5aNZrG36rth4a?dl=0 

Srl.No.|Target OS|Build Env. OS(1)|Qt version(2)|Binary Package
--------|-----------|----------|------------|------------------
1|Yosemite(10.10)|Yosemite(10.10)|5.10.1(3)|klayout-0.25.4-MacOSX-Yosemite-1-Qt5101mp.dmg
2|El Capitan(10.11) Sierra(10.12)|El Capitan(10.11)|5.11.1|klayout-0.25.4-MacOSX-ElCapitan-1-Qt5111mp.dmg
3|HighSierra(10.13)|HighSierra(10.13)|5.11.1|klayout-0.25.4-macOS-HighSierra-1-Qt5111mp.dmg

Notes:
(1) using Ruby and Python bundled with OS
(2) using packages from MacPorts
(3) looks the last available version on Yosemite

Best regards,
Kazzz

klayoutmatthias:
Hi Kazunari,

many thanks for your efforts! I have uploaded the packages to the server.

Best regards,

Matthias


Quoting Kazunari Sekigawa <notifications@github.com>:

thomaslima:
Sorry for coming here late. I've pushed a pull request for the binaries. Please check https://github.com/klayoutmatthias/klayout/pull/168. I think the builds are in order (logs are attached), but I'd appreciate help testing klayout in respective virtual machines. Thank you @klayoutmatthias and @Kazzz-S .

Kazzz-S:
## Quick Test on macOS 10.14 Mojave ##

The binary package built for macOS10.13 (HighSierra) looks work on macOS 10.14 (Mojave), too,  without any change :-). 

I've tested on a virtual machine as usual. 
Thank you for your continuous contributions: @klayoutmatthias and @thomaslima.

Just for your information,
Kazzz

![01-macosmojave-klayout0254](https://user-images.githubusercontent.com/16847255/46255899-2b1e7400-c4de-11e8-807e-2385dbd798e0.png)
![02-macosmojave-klayout0254](https://user-images.githubusercontent.com/16847255/46255901-2e196480-c4de-11e8-9c32-55186102ed59.png)
[...]

PR #180, opened 2018-10-07
Updated tools/environment to build on macOS Mojave

Kazzz-S:
@klayoutmatthias 

Attempted to build KLayout on macOS 10.14 Mojave. It was bit painful.
For the time being, it would be better not to build on Mojave since "ruby.h" is not included in the OS.
Please refer to https://github.com/Kazzz-S/klayout/issues/29

Fortunately, binary package built on HighSierra works on Mojave, too (#160).

Regards,
Kazzz


thomaslima:
Hello @Kazzz-S . I am confused, does the build work in Mojave or not?

I have found a typo in the image you compiled: it says "Ysemite" instead of "Yosemite".

Another question I have is what is the purpose of the scripts KlayoutEditor.sh and KlayoutViewer.sh? People I know have been confused by it and preferred to use the normal Klayout.app bundle, then set it to editor or viewer mode. Besides, I have changed the bundle so that "open with" works with the bundle klayout.app, and set the extension .gds, .gds2 and .oas to Klayout (see Info.plist), which wouldn't if the default app was one of the scripts.

Kazzz-S:
Hello @thomaslima,

Thank you for pointing out the typo. I'll fix it soon.

----

>  I am confused, does the build work in Mojave or not?

I've found "ruby.h" is not included in the current 10.14 macOS.
Therefore build "using OS bundled Ruby as is" fails.
With some workaround (https://github.com/Kazzz-S/klayout/issues/29), I could build in Mojave, too.
So far, the major build problem in Mojave is "missing ruby.h".  I hope the next OS release will fix this.
I have not tried other choices (e.g. Ruby from Homebrew).

----

> Another question I have is what is the purpose of the scripts KlayoutEditor.sh and KlayoutViewer.sh? 

There is a little historical reason.
Earlier (even now?), Windows binary package provided two types of launching icons; one for Editor and another for Viewer. I liked this idea. 
In my main work environment (Linux), I've prepared two icons each of which is passed different "-style" option to Qt. More precisely, with Qt4,  I'm using "plastique" for Editor; "motif" for Viewer.

I wanted to mimic this in Mac, too, several years ago. No change, since then.
If Mac users do not need this (even optionally), we can drop this in future.

Regards,
Kazzz

thomaslima:
Thanks for the explanation!

Best,
Thomas

lukasc-ubc:
I agree. I don’t think the extra Editor/Viewer scripts are necessary. Just the App suffices.

thomaslima:
That looks great to me. Thanks!

On 8 Oct 2018, at 07:25, Kazunari Sekigawa <notifications@github.com<mailto:notifications@github.com>> wrote:


[2018-10-08 19 58 11]<https://user-images.githubusercontent.com/16847255/46606173-dc548800-cb36-11e8-98f8-6da89bcd87e1.png>

How about this? Sorry for not being a good designer!
Kazzz

PR #184, opened 2018-10-09
Pymod Creating pip package

thomaslima:
- added MANIFEST file to support source distribution with `setup.py sdist`
- fixed bug that prevented db_plugins to be loaded. Explained in 
https://github.com/klayoutmatthias/klayout/commit/e7ebd88b2cdd0b3bf42888bcce46569e519aeee6
- triggering unit_tests (import_tl, import_db, import_rdb) from wheel building in travis. This should check whether db_plugins loads from pymod.


thomaslima:
We have a pip package! https://pypi.org/project/klayout/#history

PR #186, opened 2018-10-19
Pymod package splitting

thomaslima:
Refactored the pymod package so that the CPython extensions were renamed to *.core. For example, `klayout.db` is now `klayout.dbcore`. Then, I created `klayout.db` which imports all names from `klayout.dbcore`, and added the PCellDeclarationHelper, for example, which is present in `pya`.

I had to change the folder structure a little bit due to a setuptools bug https://github.com/pypa/setuptools/issues/230

I also had to create an empty folder named db_plugins inside `src/pymod/distutils_src/klayout` so that `python setup.py develop` would function properly. Now the editable install, i.e. `pip install -e .` works.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -15,7 +15,7 @@
        DEF reader did not pull vias from LEF
 * Bugfix: https://github.com/klayoutmatthias/klayout/issues/174
        Performance issue with many layers with width >1
-* Bugfix: https://github.com/klayoutmatthias/klayout/issues/175
+* Bugfix: https://github.com/klayoutmatthias/klayout/issues/176
        Painting issue with texts
 * Bugfix: https://github.com/klayoutmatthias/klayout/issues/185
        Hash values available as __hash__ standard method now
--- a/setup.py
+++ b/setup.py
@@ -54,13 +54,15 @@ and won't find them. So we need to take away the path with
 "-Wl,-soname" on Linux (see Config.link_args).
 """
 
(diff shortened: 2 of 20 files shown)


klayoutmatthias:
Hi Thomas,

I have changed the version to .dev7 which is the next one available. Some conflicts I resolved (testing is yet to do).

Matthias

thomaslima:
Sounds good. Nice work.
I am a bit confused though. Does it mean that normal qmake also builds pymod as well as setup.py? Is it supposed to build pymod because of the Qtbindings?

klayoutmatthias:
No worries, nothing wrong :-)   ... qmake-based Python modules were the first. They are still provided to supply the Python modules coming with the Debian and RPM packages. I see one more benefit: as they are integrated with the main application's build, they are part of the test suite. They basically share most of the source code with the setuptools-based distribution (again), so there is little overhead (except me having to maintain the qmake-builds and RPM/DEB packager scripts). 

BTW: there are three more small patches, so maybe you may like to pull pymod's HEAD.

Thanks,

Matthias

klayoutmatthias:
Plus one more commit ... (Python 2 compatibility)

thomaslima:
Thanks for letting me know. I am going to take a break from maintaining travis builds. It seems they're going through some changes now and every time I have to change the travis script. This instability is killing me. When pymod gets merged to master I'll try to work on CD to pypi.

PR #215, opened 2018-12-27
Bugfix: gsi method db Technology.load pointing to save

thomaslima:
Detected this bug when I tried to use it.


klayoutmatthias:
Thanks a lot for finding this. How embarrassing :-(

Matthias

PR #218, opened 2019-01-02
Azure/windows wheel builds Py 3.5,3.6,3.7 win32 and win64

### LINKED ISSUE #217: Feature request: Python Windows wheel using continuous integration
This is related to #202 and #124 .

Request 1: Since apparently there is a clean way to build layout's pymod using MSVC (https://github.com/klayoutmatthias/klayout#build-instructions-windows-msvc-2017), and one of the wheels is already uploaded to pypi (python 3.6, windows 64-bit https://pypi.org/project/klayout/0.26.0.dev8/#files), I suggest we use azure pipelines or appveyor to automatically build these wheels. It will be faster and more reliable than using our own virtual machines.

Requested python versions: 2.7, 3.5, 3.6, 3.7. Win 32- (optional) and 64-bit. 

Request 2: Have the external libraries be included in the wheel. I think it is possible, judging by bumpy and scikit packages, but it might be a pain to figure out how to do it.

Once that's done, we can release a python package klayout-0.26.0.beta that works on the main versions of mac, linux and windows.

thomaslima:
Closes #217 

## Contributions:

-  Now `pip install klayout` works for python >= 3.5 on windows!

- Building pip wheels for windows (including libcurl, expat, zlib and pthread DLLs). I had to manually patch the wheel after the build to place the DLLs in the klayout module folder. *Python versions 3.5, 3.6, 3.7*, using MSVC++ 14 in *both Win 32 and 64* architectures. 3.4 and below would require MSVC++ 10.0, but I haven't tested them.

- I uploaded all wheels to pypi under version 0.26.0-dev10, excepting cp36 x64, which was already there for some reason. See https://dev.azure.com/thomaslima/klayout/_build/results?buildId=50&view=results

Issue: I could not build a wheel for Python 2.7. It uses a very outdated C++ compiler (MSVC++ 9.0) and KLayout's source fails to compile. This is pretty much what happens: [log_10_22.zip](https://github.com/klayoutmatthias/klayout/files/2720232/log_10_22.zip) I am not sure it is worth providing python 2.7 support for windows. (Mac and Linux still ok though)

### Todo:

- [x] - Maybe build part of klayout-bits in azure as well for completeness. Right now I am using pre-built "klayout-microbits-1.0.zip" which originated from klayout-micro-1.0.zip, but without python and ruby which take up a lot of space. [klayout-microbits-1.0.zip](https://github.com/klayoutmatthias/klayout/files/2720246/klayout-microbits-1.0.zip)
- [x] - Release 0.26.0-dev11 because the cp36 dev10 version was uploaded without the external library patch.
- [x] - Figure out how to organize continuous deploy in conjunction with Travis (or should we move to azure entirely?)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,12 +5,12 @@ jobs:
     vmImage: 'vs2017-win2016' # other options: 'macOS-10.13', 'ubuntu-16.04'
   strategy:
     matrix:
-      Python27:
-        python.version: '2.7'
-      # Python35:
-      #   python.version: '3.5'
-      # Python36:
-      #   python.version: '3.6'
+      # Python27:
+      #   python.version: '2.7'
+      Python35:
+        python.version: '3.5'
+      Python36:
+        python.version: '3.6'
     maxParallel: 3
 
   steps:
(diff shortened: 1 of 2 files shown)


thomaslima:
I prepared an azure pipeline for the klayout microbits: https://github.com/klayoutmatthias/klayout_bits4msvc2017/pull/1

Download available forever from bintray:

[ Download ](https://bintray.com/lightwave-lab/klayout/klayout-microbits/1.0/link)

thomaslima:
I've done some thinking about CD, and I believe it is out of the scope of this PR. For now, we can upload packages manually as it is done right now. Later, we trigger automatic deployment as soon as there is a tagged commit on master. We just have to be diligent and only tag commits that passed a series of tests. Thus Azure can be responsible for uploading windows wheels, and travis for linux and macOS. I'll open a new issue after you review this PR about a klayout-0.26.0-beta release.

thomaslima:
BTW, builds can be found here: https://dev.azure.com/thomaslima/klayout/_build?definitionId=1

PR #250, opened 2019-03-31
Spelling errors in code and comments fixed

rubund:
Hi Matthias,

I'm working on the Debian package, and these spelling errors were discovered. Please consider applying them upstream so that we don't have to patch them when packaging.

Cheers
Ruben


klayoutmatthias:
Whow ... thanks for these fixes.  :-)

klayoutmatthias:
I admit I wasn't really careful with spelling in comments. Spell checking wasn't always in there with editors, so I did not notice.

Thanks anyway for that lot of work.

Best regards,

Matthias

PR #253, opened 2019-04-03
Continuous Deployment to PyPI

thomaslima:
- Automatically upload wheels to PyPI (configured azure and travis with my credentials)
- Stripped shared objects in linux pymod build to reduce wheel size by 90%
- Upgraded version to klayout 0.26.0.dev14
- Preview of the wheels in https://test.pypi.org/project/klayout/#files (will trigger upload to main pypi when this PR is merged)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -585,6 +585,6 @@ after_success:
   # uploading to pypi using twine
   # need TWINE_USERNAME, TWINE_PASSWORD and TWINE_REPOSITORY_URL env variable
   - if [ "$PYTHON_BUILD" = true ] || [ "$DOCKER_BUILD" = true ]; then
-      pip install -U twine;
+      pip install -U twine || sudo pip install -U twine;
       twine upload --skip-existing deploy/dist-pymod/$klayout_version/*;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -578,4 +578,13 @@ script:
     fi
 
 after_success:
+  # upload to dropbox
+  # need DROPBOX_OAUTH_BEARER environment variable
   - dropbox-deployment
+
+  # uploading to pypi using twine
+  # need TWINE_USERNAME, TWINE_PASSWORD and TWINE_REPOSITORY_URL env variable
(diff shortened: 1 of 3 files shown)


thomaslima:
Matthias, right now deployment to PyPI will happen at any commit on any branch. So the first commit with an updated version will trigger a new version in pypi. When, for example, you upload a Windows wheel with a new version, pip install klayout on Linux or Mac starts failing because there are no wheels for that platform in the new version. 

The solution for this is to upload all available wheels at once when a new version is available. We should come up with a release strategy, for example deploy to pypi for commits on pypi branch only, or only for tagged commits etc.

klayoutmatthias:
Hi Thomas,

many thanks for this patch. 

Actually I thought that pip only disables overwriting the same file, so a release can incrementally be completed. If you're right, a release strategy might be to collect all wheels on a dropbox and release it at once, right?

Best regards,

Matthias

klayoutmatthias:
Hi Thomas,

one more question: in order to use twine I had to give it configuration file with my credentials. I don't want to publish this, so I wonder how credentials are managed in Travis/Azure. Do I need to supply them somewhere?

Thanks,

Matthias

thomaslima:
A release can incrementally be completed, but while we're providing wheels one by one, pip install may fail for some users.

In the past, that's what I did. I collected all wheels from azure and travis to my dropbox, then uploaded them all at once.

Now, azure and travis will upload straight to pypi instead of my dropbox. This assumes that the build succeeds in all platforms. If it succeeds in one and fails in another, we have the same lag problem. That's why we should only deploy bug-free commits.

My strategy is to use secret environment variables. I defined my TWINE_USERNAME and TWINE_PASSWORD in both azure and travis, so it will get uploaded using my credentials. Feel free to change to your credentials, but it's honestly not necessary. I'll tell you in advance if I'm not able to maintain these wheels.

dev14 will be uploaded to pypi when the following builds finish:
- https://travis-ci.org/KLayout/klayout/builds/515283103 (Mac and Linux)
- https://dev.azure.com/klayout/klayout/_build/results?buildId=7 (Windows py>=3.5)

PR #254, opened 2019-04-03
A few more spelling fixes

rubund:
Please check if the fixes are correct.

Also lintian complains about a number of:
 "allow to"/ "allows to"
which apparently are supposed to be:
 "allow one to" / "allows one to"

I have not corrected those since I am not sure if it is correct in all cases.

Cheers.


klayoutmatthias:
Oh yes ... bad english :-(   I used to say "allows to check ...", but in fact it's "allows checking" or "allows one to check". I'll try to find those places and fix it.

Best regards,

Matthias

Issue #376, opened 2019-10-08
Parameters UI callback invoking 


rizoschrist-prime:
Callback invoked when a PUI item loses its focus or its status changed.

klayoutmatthias:
Which callback do you mean? Any code example?

klayoutmatthias:
I see your PR which is OK ... but could you give some motivation for this fix?

PR #377, opened 2019-10-08
PCell callbacks invoked on value change.

rizoschrist-prime:


klayoutmatthias:
If you mention the bug number in the commit message (like fixed #376), PR and issue get linked. This way it's easier for me to identify PR and ticket.

src/edt/edt/edtPCellParametersPage.h
```diff
@@ -111,7 +111,7 @@ Q_OBJECT
   void set_parameters (const  std::vector<tl::Variant> &values);
 
 public slots:
-  void activated (int);
+  void activated (int i=0);
```
klayoutmatthias: activated(int) can be replaced in all places by "activated()" simply. Signals can be connected to slots with less parameters.

rizoschrist-prime: Hi Matthias, thanks for the review.

As for the purpose of this feature. I have tried to reach you via DM from KLayout forum to ask you for some hints about the changes I wanted to integrate, please it would be much of a help to check them.

Specifically for the purpose of this change, it is easy for someone to "Apply" a corrupted geometry, without continuously checking the input values. When a UI item loses its focus, or its status changed, 'get_parameters' should be invoked to continuously check the status of all parameters after the possible change. When the user finishes tuning the parameters and is ready to "apply" a layout geometry, the parameters should have already been checked and correspond to a valid geometry, to avoid any corrupted structures.

I thought that it is best practice the signature of a signal to match the signature of the receiving slot, although it is acceptable the slot to have fewer arguments from the activated signal. Boolean signal 'stateChanged' has 'int' in its signature, so I wanted to cover both 'editingFinished' and 'stateChanged' with this implementation.

Should I make a new PR with your proposal, or it is possible to push the changes to the current request?

PR #451, opened 2019-12-15
Fixed #447 (netlist extraction failed on meander test case)

### LINKED ISSUE #447: Net extraction bug on attached test case
The net extraction on layer 1 of the attached test case renders three nets instead of a single one. A problem with the array instance handling probably.

[meander.gds.gz](https://github.com/KLayout/klayout/files/3941084/meander.gds.gz)

klayoutmatthias:


klayoutmatthias:
Will be done through dvb branch in other PR - too many interactions.

Issue #448, opened 2019-12-11
LVSDB database not readable with device mapping


klayoutmatthias:
When using device mapping in LVS, the LVSDB written cannot be read because the device class for the reference netlist can't be found.

Solution: include device classes also in the reference netlist.

klayoutmatthias:
This only happens in the short format.

PR #452, opened 2019-12-15
#447 and #448 fix (meander case extraction bug, LVSDB writing box)

klayoutmatthias:


klayoutmatthias:
Includes #447 and #448 fix

PR #562, opened 2020-05-22
Add klayout.metainfo.xml metadata file

willsALMANJ:
This PR adds a minimal AppStream metadata file (see [documentation here](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html)) which helps with displaying the application in software centers. Whether you want to do use it is up to you. I am providing it because I want to package KLayout as a flatpak and it is recommended to offer this file upstream.


willsALMANJ:
By the way, do you run any script when you make a new release? It seems like making a new version requires at least updating version.sh and making a new tag. I was wondering if there were a way to make it easy to add new version numbers to the "versions" section of the xml file when a new version is released.

PR #635, opened 2020-09-04
building wheel for python3.8

### LINKED ISSUE #633: pip install klayout on python>=3.8
@thomaslima 

How can we pip install klayout on python>=3.8?

joamatab:
closes #633

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,36 @@ branches:
 matrix:
   include:
     # python manylinux packages
+    - name: "cp38-cp38m-manylinux1_x86_64.whl"
+      os: linux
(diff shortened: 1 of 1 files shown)


klayoutmatthias:
@joamatab  Many thanks for providing this patch!

I have merged it into the 0.26 branch. My understanding is that this should build pypi/klayout for 0.26.7 ... let's see what's going to happen.

klayoutmatthias:
Ok, it's happening not yet, because we made it act only on new "vx.yy.z" tags. I'll create a tag like "v0.26.7-1". We should have a build then.

PR #668, opened 2020-11-04
Add -lexecinfo for BSD

Partmedia:
FreeBSD and NetBSD require -lexecinfo for the backtrace() functions.


Partmedia:
Actually, something's not quite right; I'm still debugging. The LIBS are correct, but there seems to be something inconsistent between when I built with 'LIBS' and when I built with '*bsd*:LIBS' that is making the build fail again for me. I'll update this once I figure out what's going on.

Partmedia:
I've updated the pull request. I think this is correct now.

PR #672, opened 2020-11-10
build and upload wheels for python3.9

joamatab:

PR #676, opened 2020-11-12
build py39 wheels

joamatab:


thomaslima:
Dear @joamatab 

The rebase was not done correctly. I expected to see one commit built on top of the branch called "0.26". And the pull request needs to point to the branch 0.26, not master.

klayoutmatthias:
I have merged PR #672 which was for master. If I understood it correctly, this PR is supposed to be for 0.26, right?

But somehow the rebase isn't working. I usually "rebase" such small PRs through cherry picking. I can do so if you like. We do not need another PR for this purpose.

Matthias

thomaslima:
Ideally @joamatab should rebase his branch joamatab:py39_klayout0.26 to klayout:0.26 and then this PR will update itself. It's cleaner this way.

I still don't quite understand exactly your git repo model. Because the 0.26 branch has diverged so far from master, I am assuming that the master branch is not stable and 0.26 is the one that will release the next version. That's why it does make sense to send this change to 0.26.

klayoutmatthias:
I'm being too creative :) Hence master has deviated so much. Frankly I have not found the time yet to finalize 0.27 with the paperwork (documentation etc.) and some proper testing. It should be out already.

The original PR #672 was just a single file so I manually cherry picked it into 0.26. I have create #679 for this patch of 0.26. If that's correct we may dismiss this PR and merge #679 instead.

Best regards,

Matthias

thomaslima:
My point is that if we will merge 0.26 into master prior to the next release based off the changes in master, there is no need to duplicate commits across different branches. I suggest that we just wait it out.

Off topic:
It would be nice to have a default branch that is always stable, and which gets merged onto a release branch like 0.26 at the time of release. Then there could be a development branch for new ideas that would be merged to the stable one before any next major version. Right now I see master as the development branch and 0.26 with a stable and release dual role. If you agree with something like this, all it would take is to create a new branch called stable off of 0.26 to which all harmless PRs would be merged into. Then occasionally you could merge this into master. And also merge into 0.26 when a new minor version is released. This organization would eliminate the need to cherry pick.

________________________________

joamatab:
closing this old PR

Issue #660, opened 2020-10-23
macbuild's heavyweight homebrew python build is not patching pip correctly


thomaslima:
First of all, sorry for being away all this long. It's been a very tough year for me.

I appreciate the new routines present for building klayout with homebrew's binaries for python. I was looking at what it takes to create a brew formula to build klayout's application against the most recent qt and python. It is surprisingly straightforward to create such a formula, actually, and it might be a more convenient way to deploy klayout for experimental users like me. I think it could also automate make the mac-brew-based CI builds more robust.

Anyway, the new versions of pip changed the internal location of the code for installing packaged. [They really don't want pip to be used as a module.](https://pip.pypa.io/en/stable/user_guide/#id9) I.e. running `pip.main(['install', 'numpy'])` from within klayout. The reason is that they want to quit the python process after being done... I knew that but took the risks anyway, after all I can always reboot klayout after installing packages.

That aside, pip install is currently putting packages in `~/Library/Python/3.8/lib/python/site-packages` instead of `klayout.app/Contents/MacOS/python`, which was the original intention. I think this is because the `sys.path` seen from within klayout contains the home Library before the klayout.app's location. For example:

```
>>> sys.path
['/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python38.zip', '/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8', '/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/lib-dynload', '/private/tmp/klayout-test-20201022-16428-1yr8l4j/klayout-homebrew-formula/.brew_home/Library/Python/3.8/lib/python/site-packages', '/usr/local/opt/python@3.8/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages', '/private/tmp/klayout-test-20201022-16428-1yr8l4j/klayout-homebrew-formula/HW-qt5Brew.pkg.macos-Catalina-release-RsysPhb38/klayout.app/Contents/MacOS/python', '/private/tmp/klayout-test-20201022-16428-1yr8l4j/klayout-homebrew-formula/HW-qt5Brew.pkg.macos-Catalina-release-RsysPhb38']
```

[...]

PR #687, opened 2020-11-28
Better python integration with klayout

thomaslima:
This PR is designed to address #660 . So far, I have done the following changes to the macbuild installation:
- shipping a python binary alongside klayout, which imports the same libraries that the python interpreter within klayout does
- within that instance, I replaced `sys.executable` with that python binary rather than the klayout binary.

The motivation for this is to allow packages to install python packages with pip install by using a thread-safe `subprocess` call, e.g.:
```python
subprocess.check_call([sys.executable, '-m', 'pip', 'install', 'numpy'])
```

With this PR, in the "heavyweight" HW distribution, this command places numpy into the `klayout.app`. It also does not crash klayout's python interpreter in case this fails.


klayoutmatthias:
I think a lot happened in that respect ... can I close this PR?

Thanks,

Matthias

klayoutmatthias:
Hi Thomas,

I'll close this PR - I think a lot has been done in that direction since then :)
If you want to reopen, let me know.

Best regards,

Matthias

PR #785, opened 2021-05-01
Changes in the build system for Mac

Kazzz-S:
Changed and updated the build system for Mac to support the Apple Silicon chip plus Big Sur.


Kazzz-S:
I'll cherry-pick these changes into 0.27 and test, then send another pull request later.

Kazzz-S

PR #804, opened 2021-05-21
Fix documentation typos

nikosavola:
This fixes a lot of typos systemically for API documentation (I used https://github.com/myint/scspell)

This started as I noticed a couple of typos while reading the API docs, then I started noticing  more and more. Eventually, I chose to systemically go through the `gsiDecl*.cc` files and fix the typos.


klayoutmatthias:
Thanks a lot for your efforts. I appreciate that!

PR #820, opened 2021-06-03
Fix misspellings in laybasic.

EugeneZelenko:


klayoutmatthias:
So much misspelling ... how embarrassing ...

Thanks for fixing that! :)

Matthias

klayoutmatthias:
BTW: how did you find this? Do you have a source code spell checker?

EugeneZelenko:
I use Code Spell Checker extension in Visual Studio Code. It's great tool for everyday work (spelling mistakes highlighted in same way as static analysis tools like Flake8, PyLint or Clang-tidy), but it takes time to walk through legacy code.

klayoutmatthias:
I'm using SonarQube community editing for static analysis, but have not found a spell checking plugin yet. Would be much easier that way ...

klayoutmatthias:
For the IDE I'm using QtCreator. If there is a spell checking option I have not found it either ...

EugeneZelenko:
Only limitation of Code Spell Check that it doesn't allow to ignore words in same context (for example in code constructs or XML tags), so it's a little bit annoying to ignore same thing again and again :-) But what is right for code is not necessarily right for strings and comments.

PR #932, opened 2021-11-10
Fix format-security error

xctan:
This pr tries to fix format-security errors when klayout is compiled with `-Werror=format-security` flag.


klayoutmatthias:
Thanks for this PR, but I cannot accept it.

Problem is that file is auto-generated from Qt headers and must not be edited (as the comment at the top says).

The root problem is an incomplete mapping of the Qt C++ API (the "..." are skipped by the translator script). This is harder to fix than through this patch.

I have created a ticket for the real patch: https://github.com/KLayout/klayout/issues/933

Issue #1069, opened 2022-04-28
How to make `pip install klayout` in macOs faster?


joamatab:
some macOs gdsfactory users have reported that `pip install klayout` takes 30min for mac

How could we make it faster?

See the issue here 
https://github.com/gdsfactory/gdsfactory/runs/6214659865?check_suite_focus=true

is there any version for klayout that could make it faster?

@klayoutmatthias 
@thomaslima 
@flaporte
@tvt173
@lukasc-ubc

thomaslima:
Yes. I have this in my backlog. We need to create a wheel for the newer macOS versions and upload to pypi. Right now it is compiling from source. Travis stopped their easy free open source service; last I checked they wanted a credit card to offer free builds. I will check GitHub actions (didn't know they offered macOS builds).

joamatab:
Hi Thomas,

yes, it takes 30min to compile from source :)

we can set up github actions for this
https://github.com/KLayout/klayout/pull/1070

let me know what you think
Joaquin

klayoutmatthias:
@joamatab: @thomaslima is correct. Travis has resource restrictions now that basically make it impossible to deploy KLayout packages.

But I see the GitHub actions on the original fork failing and they are only executed on Linux. So is that functional? Is MacOS available with a free plan?

Plus is there going to be active maintenance? I personally do not have the resources for continuous support of another CI queue.

Matthias

joamatab:
i think is not working yet
https://github.com/joamatab/klayout/runs/6220334022?check_suite_focus=true

```
   g++: error: build/temp.linux-x86_64-3.8/src/tl/tl/tlHttpStreamCurl.o: No such file or directory
    error: command 'g++' failed with exit status 1
```

once is working, we can deploy wheels that work for all os (linux, macos, windows) as they are available for free in github actions

another great thing would be to have a conda package for klayout

@proppy

klayoutmatthias:
From what I see the job you're proposing is only pulling the sources from PyPI and generating wheels for particular platforms on a regular basis. So it's not really CI/CD.

So if it's not connected to KLayout's sources here and as this job is not tied to branches or pull requests here I'd prefer to have it separately. I propose to set up a new repository - i.e. klayout/pypi-buildbot - and put the GitHub actions there.

joamatab:
Hi Matthias,

[...]

PR #1070, opened 2022-04-28
add ci/cd for github actions

joamatab:
add ci/cd for uploading pypi wheels for macOs, windows and Linux


klayoutmatthias:
Ah, you're right. I stand corrected.

I have a few remarks though:
* I don't see a dependency of the twine upload on the "v*" tag. You must not upload other builds as you cannot delete files on PyPI, specifically not "master". Ideally builds are done on master and PRs but twine deployment only happens in "v" tags or "on release".
* I like the manylinux scheme. Any chance to port it? I'm afraid just offering Ubuntu will not be enough.
* Do GitHub actions cache the builds or will every master commit trigger a full rebuild? That will be costly in terms of resources.

I need to say, that I'm actually quite happy to have reduced my dependency on "free" services. These in fact generate some considerable workload by constantly upgrading their environments, certificates, authentication schemes etc. For example I moved away from Azure with Python deployment after they had removed all their MSVC2017 support (MS C++ compilers are magic, so MSVC2019 did not work out of the box ...). To my taste these services do not offer enough configuration options and stability for C++ deployment. I do not expect much more from GitHub actions.

thomaslima:
Dear @joamatab ,

I have reworked this PR to use cibuildwheel, as mentioned in #1069 . Can you review and merge my commits from https://github.com/thomaslima/klayout/tree/issue-1070 into your branch? There's a single workflow file now that does all the ubuntu manylinux/musllinux builds plus macOS builds (with cache between builds). The advantage is that cibuildwheel is maintained by pypa so it will take little work to update to future python versions.

@klayoutmatthias I have already created a pypi_password and test_pypi_password secret token. Right now, it is configured to upload to test pypi at every commit. Please uncomment lines 82 and 96 and comment line 94 in `build.yml` before/after merge to change the target repository to the real pypi one.

Thanks,
Thomas

joamatab:
Awesome! thank you Thomas, I merged your commits

PR #1103, opened 2022-06-23
Add setting to disable Save Needed dialog box

### LINKED ISSUE #1102: Add setting to disable Save Needed dialog box
I mainly generate/edit layouts with Python and don't really need to save layouts. For this kind of workflow, I find the _Save Needed_ dialog box a bit unnecessary.

I suggest that there should be an option in the settings to always allow _Exit Without Saving_. This should be off by default.

image

nikosavola:
This PR adds a setting to disable the _Save Needed_ dialog box.

#### Additionally
- a random quotation mark here is removed 
image
- a dot added to finish the sentence in _Watch Layout Files_

Closes #1102

### DIFF
--- a/src/lay/lay/MainConfigPage7.ui
+++ b/src/lay/lay/MainConfigPage7.ui
@@ -57,7 +57,7 @@
       <item row="0" column="0" colspan="2">
        <widget class="QLabel" name="label">
         <property name="text">
-         <string>If this option is enabled, the program will detect changes in layout files and ask whether to reload them</string>
+         <string>If this option is enabled, the program will detect changes in layout files and ask whether to reload them.</string>
         </property>
         <property name="wordWrap">
          <bool>true</bool>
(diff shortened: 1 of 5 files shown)


klayoutmatthias:
Thanks for the patch, but this will not work.

You're not checking against the value of the option, but against the string containing the configuration option name in `layMainWindow.cc` (`cfg_always_exit_without_saving`). This probably always exits without asking.

The correct way is to introduce a boolean member in `lay::MainWindow` (e.g. `bool m_always_exit_without_saving`), initialize it with false in the constructor, set its value from the `cfg_always_exit_without_saving` configuration option in `lay::MainWindow::configure` and use this boolean in the if condition.

Matthias

nikosavola:
> Thanks for the patch, but this will not work.
> 
> You're not checking against the value of the option, but against the string containing the configuration option name in `layMainWindow.cc` (`cfg_always_exit_without_saving`). This probably always exits without asking.
> 
> The correct way is to introduce a boolean member in `lay::MainWindow` (e.g. `bool m_always_exit_without_saving`), initialize it with false in the constructor, set its value from the `cfg_always_exit_without_saving` configuration option in `lay::MainWindow::configure` and use this boolean in the if condition.
> 
> Matthias

Hello Matthias. Thanks for the feedback.

I updated the code, is it now correct? I'm having problems with the compiler so it's a bit hard to test 😅

klayoutmatthias:
Yes, that's fine now! :)

PR #1155, opened 2022-09-15
Minor improvements to klayout python package

thomaslima:
Hello. I have been in the process of refactoring some of the code in https://github.com/lightwave-lab/zeropdk, and in the process I have made a few improvements to the database API. I also have fixed a few issues with the stub files (e.g. distinguishing Sequence from Iterators). Type hinting is now top notch and helped me detect bugs I hadn't foreseen.

- Implemented __deepcopy__ for Point/Vector objects.
- Changed exception for wrong argument type in method call from RuntimeError to TypeError.
- Implemented p - v = p operation, where p:Point and v: Vector, matching p + v = p
- Distinguished between Iterator and Sequence in Python stubs. 
  - Methods like object.each() typically returns an iterator, which can't be indexed in python. 
  - Methods like object.merge() returns an actual sequence/list-like object that can be indexed like list[0].
- For operators such as +, -, *, | etc., return NotImplemented when the operand is unknown. 
  - For example p * x, where p: Point and x: Unknown, would yield TypeError if x is not a float. 
  - Now, it returns NotImplemented, so that python calls x.__rmul__(p). 
  - This is useful when x is a numpy array. The example above returns an ndarray of points.
- Disabled __rmul__ for Trans objects. (Not commutative)
- Enabled GSI_ALIAS_INSPECT to activate __repr__.
- Implemented __setstate__ and __getstate__ for objects that had `from_s`, `to_s` and `assign` defined. Useful for pickling and serialization, as well as deep copying.
- Implemented __copy__ for objects with dup() methods.
- Annotating return type vector as List instead of Sequence.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,5 @@ dist/
 # Macos artifacts
 *.dmg
 *.dmg.md5
+.DS_Store
+
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,3 @@
-
 """
 
 KLayout standalone Python module setup script
@@ -54,6 +53,7 @@ and won't find them. So we need to take away the path with
 "-Wl,-soname" on Linux (see Config.link_args).
 """
 
+from typing import List
 from setuptools import setup, Distribution, find_packages
 from setuptools.extension import Extension, Library
 import glob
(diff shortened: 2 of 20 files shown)


klayoutmatthias:
Hi Thomas,

thanks for the patch. It actually isn't "minor" :)

A few comments:
* You're right about the non-commutative nature of transformation multiplication. But I don't like to see a name match - I'll never remember to add more classes here if required. I'd rather go for an annotation (e.g. of "*!" as "non-commutative *"), but as of now we can leave it like this. I'll change it when porting the patch to master.
* I see that "Point-like" objects get a special handling through "convert_type_error_to_not_implemented". Why not generalize - i.e. mapping all non-implemented operators to "NotImplemented" for all classes?
* In general "dup" is always a deep copy, so does it make sense to map "dup" to "__deepcopy__" always?

Best regards,

Matthias

thomaslima:
Hi Matthias,

Thanks for reviewing my PR! Sorry for the late reply.

> * You're right about the non-commutative nature of transformation multiplication. But I don't like to see a name match - I'll never remember to add more classes here if required. I'd rather go for an annotation (e.g. of "*!" as "non-commutative *"), but as of now we can leave it like this. I'll change it when porting the patch to master.
I think that having `rmul` is a "nice to have" feature. In python, it's not usually expected from libraries. I would prefer if you picked an annotation for commutative operations instead. You can take your time with this, I think it pays to get it right for both master and 0.27.
> * I see that "Point-like" objects get a special handling through "convert_type_error_to_not_implemented". Why not generalize - i.e. mapping all non-implemented operators to "NotImplemented" for all classes?
You are right. I was lazy and only used it in the classes I use most often. `Box`, for example, is missing. I'll patch it.
> * In general "dup" is always a deep copy, so does it make sense to map "dup" to "**deepcopy**" always?
I didn't know dup always did a deep copy, especially for the `Instance` and the `ParentInstArray` classes. Could you check [this link](https://docs.python.org/3.7/library/copy.html) for the requirements for implementing the `__deepcopy__` memo argument? Namely, is there any chance that objects contain references to itself, creating a recursive copying issue?

Thanks again!

klayoutmatthias:
Sorry for taking so long for the merge and thanks for the effort you spent! 

Matthias

PR #1160, opened 2022-09-23
Use enhanced tab widget for file tabs to easily navigate through larg…

EugeneZelenko:
…e number of opened files.

### DIFF
--- a/src/lay/lay/lay.pro
+++ b/src/lay/lay/lay.pro
@@ -11,6 +11,7 @@ HEADERS = \
     layClipDialog.h \
     layControlWidgetStack.h \
     layCrashMessage.h \
+    layEnhancedTabWidget.h \
     layFillDialog.h \
     layGSIHelpProvider.h \
     layHelpAboutDialog.h \
@@ -121,6 +122,7 @@ SOURCES = \
     layClipDialog.cc \
     layControlWidgetStack.cc \
     layCrashMessage.cc \
+    layEnhancedTabWidget.cc \
     layFillDialog.cc \
     layGSIHelpProvider.cc \
     layHelpAboutDialog.cc \
--- a/src/lay/lay/layEnhancedTabWidget.cc
+++ b/src/lay/lay/layEnhancedTabWidget.cc
@@ -1,0 +1,103 @@
+
+/*
+
+  KLayout Layout Viewer
+  Copyright (C) 2006-2022 Matthias Koefferlein
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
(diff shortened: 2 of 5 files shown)


EugeneZelenko:
Only missing piece is icon. Looks like KLayout is in the middle of refactoring icons handling, so I'll need guidance on that. Icon I have in mind: https://fonts.gstatic.com/s/i/short-term/release/materialsymbolsoutlined/menu/default/48px.svg.

klayoutmatthias:
Thanks a lot for this PR. I'll take a look. 

You're right, there is refactoring going on with the goal to support High-DPI icons. I will produce SVG for the new icon as well.

Matthias

klayoutmatthias:
Thanks again for providing this PR. I really like this feature :)

However, after taking a look I have some modification proposals. Specifically, I'm slightly worried about the implications of switching to QTabWidget. I made an alternative proposal in this branch: [enhanced-tab-widget
](https://github.com/KLayout/klayout/tree/enhanced-tab-widget). I cannot modify your PR, but maybe you want to take a look.

For the rationale see here: https://github.com/KLayout/klayout/commit/ab9fe57ec3363554c435163db7771ab00029e7a7

Best regards,

Matthias

EugeneZelenko:
Your branch looks OK for me. `QTabBar/QTabWidget` switch was not trivial because of life-time management, but somehow `QTabWidget` worked for me :-)

I noticed that you added context menu for tabs, so it may be useful also to add `Close to the Left/Right` like Visual Studio Code of Firefox. `Clone` command also seems to be good fit to context menu.

klayoutmatthias:
Makes sense, yes.

I have added these features (also for the macro editor tabs except for "clone") to the branch if you like to take a look.

Matthias

EugeneZelenko:
Just build and tested your branch. Looks OK for me, except there is different between KLayout and VS Code/Firefox how to treat context menu. In KLayout currently selected tab is base one, but in VS Code/Firefox base tab is where mouse was right-clicked (i.e. `QTabBar::tabAt`). Theoretically right click could be made outside bars, so `-1` return value should be handled too (no menu displayed).

klayoutmatthias:
I see ... it's not trivial to detect the tab the mouse clicked appeared at, but I'll try.

EugeneZelenko:
I may be just change current tab on right-click?

klayoutmatthias:
That was easier than I thought. I have updated the branch accordingly. I cannot avoid however that the selected tab switches to the one the mouse points at. But I think that makes sense considering that otherwise the selected tab might become invisible.

EugeneZelenko:
I tried current version of branch and it works fine. Thank you!

klayoutmatthias:
Thanks! I merged this PR.

EugeneZelenko:
Thank you!

PR #1193, opened 2022-11-23
Issue 1189 High memory footprint for DRC deck

klayoutmatthias:


gatecat:
Thanks, this is working well for me locally. I'm wondering if it would be possible to get a release with this fix in too?

klayoutmatthias:
Yes, there is. 

I will release an additional 0.27.13 without that fix (stable branch patch) and then 0.28 (target is in two weeks) with this fix.

Best regards,

Matthias

PR #1205, opened 2022-12-04
MacPorts' Ruby version was updated from 3.1.2 to 3.1.3

Kazzz-S:
MacPorts' Ruby version update has been incorporated.


Kazzz-S:
Let me withdraw this PR.
I'm planning to issue another one, including this change.

PR #1222, opened 2022-12-12
Fixes segfault on some linux systems

QuantamHD:
The static order initialization seems to be more present throughout Klayout and might cause additional issues, but this fix stopped the segfault on my machine. When compiling with asan there are a number or other issues that could be addressed, but it's not clear if they're real problems.

dbLayout change occured when using the python API to read a DEF layout. When the layout is updated the code uses a const iterator which implicitly wraps the non-const iterator. In the loop the end() is implicitly wrapped, which tries to dereference a nullptr. My fix checks to ensure that the non-const iterator is not null before trying blindly dereference it.

### DIFF
--- a/src/db/db/dbLayout.h
+++ b/src/db/db/dbLayout.h
@@ -192,8 +192,13 @@ public:
    *  @brief Default constructor
    */
   cell_list_const_iterator (cell_list_iterator<cell_type> iter)
-    : mp_cell (& (*iter))
-  { }
+  { 
+    if (iter == cell_list_iterator<cell_type>()) {
+      mp_cell = nullptr;
+    } else {
+      mp_cell = &(*iter);
+    }
+  }
 
   /**
    *  @brief Equality
--- a/src/tl/tl/tlClassRegistry.cc
+++ b/src/tl/tl/tlClassRegistry.cc
@@ -23,35 +23,36 @@
  
 #include "tlClassRegistry.h"
 
-#include <map>
+#include <unordered_map>
+#include <typeinfo>
+#include <typeindex>
 
 namespace tl
 {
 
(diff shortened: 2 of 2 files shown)


klayoutmatthias:
Could you please raise an issue instead of proposing a PR? Specifically indicate which systems the problem arises on?

My preferred way of fixing the first issue was "iter.operator->()" or simply copy the pointer, but I rather like to understand how a default iterator comes into play.

I had a lot of trouble with type info pointers on Windows compilers and some gcc versions, so I am no longer sure what to rely on here. Maybe std::type_index actually is a solution but I want to see if Microsoft has the same understanding than other compilers.

Matthias

klayoutmatthias:
Never mind. I merged this commit to the wip branch. Let's see why Microsoft says.

QuantamHD:
To give some context on the system it's [GLinux](https://en.wikipedia.org/wiki/GLinux). Which is Google's in house distro based on Debian.

The other constraint we have is that all of our software gets built from source in Bazel. Part of that build system is statically linking all of your dependencies. We rewrite all of our dependencies build scripts in Bazel, and then statically link them. That might be why some of these errors show up in our build, but not Klayout's.

QuantamHD:
Looks like windows was happy with the changes. It looks like it passed in the wip branch

QuantamHD:
@klayoutmatthias Is this PR considered closed since you moved it into the WIP branch?

klayoutmatthias:
Kind of .. the CI does by far not cover everything. It will only build Python modules. I am enabling MSVC CI on my own hardware now and will close this CR when I merge the wip branch (looks good, so this is going to happen soon).

Matthias

PR #1227, opened 2022-12-21
src/unit_tests: do not link docs/icons with no qt

proppy:
without this the `without-qt` builds fails with:
```
ld: cannot find -lklayout_icons: No such file or directory
```


klayoutmatthias:
Thanks for this PR. There was a race and I had already fixed it. 

I'll close it for now. Let me know when there is another issue.

Thanks,

Matthias

PR #1229, opened 2022-12-21
Add option to enable 'Show parameter names' by default

### LINKED ISSUE #1228: Add option to have Show parameter names on by default
The _Show parameter names_ is something very useful I didn't think I would need. I would prefer to have it always on, so I suggest adding an option somewhere in Setup to have it on by default. The default behaviour for this should still stay as it is now, i.e., the parameter names are not shown.

nikosavola:
This PR adds an option to enable 'Show parameter names' by default. This option itself should be off by default

Not sure what the location for this should be in the setup, so I put it to General for now.

Drafted before tested to work.

Closes #1228

### DIFF
--- a/src/edt/edt/edtPCellParametersPage.cc
+++ b/src/edt/edt/edtPCellParametersPage.cc
@@ -227,7 +227,8 @@ PCellParametersPage::init ()
 
   mp_show_parameter_names_cb = new QCheckBox (this);
   mp_show_parameter_names_cb->setText (tr ("Show parameter names"));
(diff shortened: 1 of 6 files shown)


klayoutmatthias:
Thanks for this PR.

But honestly, I prefer the solution of https://github.com/KLayout/klayout/pull/1231. This does not introduce a new configuration parameter (which people tend not to find).

nikosavola:
Superseded by https://github.com/KLayout/klayout/pull/1231

PR #1251, opened 2023-01-15
New action: screenshot to clipboard

htherrien:


klayoutmatthias:
Makes sense.

I felt like there are so many screenshot apps around now, no one would need that feature, but it's handy.

Thanks for this PR.

Matthias

PR #1270, opened 2023-01-31
fix setup.py: py.typed not being distributed with wheel

### LINKED ISSUE #1269: py.typed is not included in the wheels
The py.typed file is currently only distributed in the sdist but not in the built wheels. I am not sure how to fix this, I have tested locally but the wheel seems to ignore the includes.

I suspect it is something related to this https://stackoverflow.com/questions/24347450/how-do-you-add-additional-files-to-a-wheel , i.e. MANIFEST.in is not respected anymore

sebastian-goeldi:
This should fix #1269

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -905,7 +905,9 @@ lay = Extension(config.root + '.laycore',
 # Core setup function
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


Ealameda31:
LGTM 👍

Issue #1265, opened 2023-01-25
Mac OSX, Python: import klayout.dbcore, missing libpng16.16.dylib


lukasc-ubc:
@Kazzz-S 

The KLayout Python package is referencing a library "libpng16.16.dylib" that is not installed on my system. I understand that it can be installed by brew.  Perhaps this library could be bundled as part of the "pip install" klayout package?

Thank you

```
(base) iMac-Hickup:anaconda3 lukasc$ pip install klayout
Collecting klayout
  Downloading klayout-0.28.3-cp39-cp39-macosx_10_9_x86_64.whl (16.7 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 16.7/16.7 MB 10.4 MB/s eta 0:00:00
Installing collected packages: klayout
Successfully installed klayout-0.28.3
(base) iMac-Hickup:anaconda3 lukasc$ python 
Python 3.9.13 (main, Aug 25 2022, 18:29:29) 
[Clang 12.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
im>>> import klayout.dbcore
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: dlopen(/Users/lukasc/opt/anaconda3/lib/python3.9/site-packages/klayout/dbcore.cpython-39-darwin.so, 2): Library not loaded: /usr/local/opt/libpng/lib/libpng16.16.dylib
  Referenced from: /Users/lukasc/opt/anaconda3/lib/python3.9/site-packages/klayout/lib_tl.cpython-39-darwin.so
  Reason: image not found

```

Kazzz-S:
Hi @lukasc-ubc,

```
(base) iMac-Hickup:anaconda3 lukasc$ pip install klayout
Collecting klayout
  Downloading klayout-0.28.3-cp39-cp39-macosx_10_9_x86_64.whl (16.7 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 16.7/16.7 MB 10.4 MB/s eta 0:00:00
Installing collected packages: klayout
Successfully installed klayout-0.28.3
(base) iMac-Hickup:anaconda3 lukasc$ python
Python 3.9.13 (main, Aug 25 2022, 18:29:29)
[Clang 12.0.0 ] :: Anaconda, Inc. on darwin
[...]

PR #1288, opened 2023-02-11
Solution for Issue 1265 - shipping libpng with macos wheels + autodetecting libpng when installing from source

thomaslima:
Fixes issue #1265 

Changelog:
- Bugfix: %GITHUB%/issues/1275 libpng was missing in macos installations.

Details:
- setup.py now looks for `libpng-config` in macos installations.
- in macos wheels, the "repair wheel" step was reinstated. it uses the `delocate` project which looks for non-system libraries and patches the wheel before uploading to pypi. this project is well maintained by pypa, so it should be stable.
- setup.py can be easily set up so that it can be installed gracefully without libpng (@klayoutmatthias decision)

Missing:
- test case for a proper use of the png library. (optional?)


klayoutmatthias:
Very good ... many thanks for this patch!

PR #1290, opened 2023-02-13
using versioneer for klayout.__version__

### LINKED ISSUE #1271: Access klayout __version__ when loaded as python module
It would be convenient for us if we could access the version string of the klayout module that is in use. A lot of python modules provide access through the use of `__version__` member, I think this could be useful for klayout as well.

This might already be available and I just didn't read the klayout docs carefully enough :D But as far as I'm aware there is only the `pya.Application.version` member, which is not available when klayout is imported as a python module without running the klayout executable.

thomaslima:
Closes #1271

- I think this is a good solution to the issue. It adds a dependency at build time (versioneer).
- versioneer adds a _version.py next to __init__.py, which is populated at build time
- if installation is editable, it dynamically computes a version from GitHub.


thomaslima:
@klayoutmatthias versioneer apparently requires python >= 3.7. That's why the tests are failing. In their defense, python 3.6 reached end of life: https://endoflife.date/python

PR #1292, opened 2023-02-13
added test for issue 1265 (writing and reading pngs)

thomaslima:
Adding a test recommended by @sebastian-goeldi . Sorry for the spammed PRs. Related to issue #1265 

WARNING: This will compare a png screenshot produced from t10.gds against a pre-recorded one. The test will fail if any presentation style of LayoutView changes in the future.

### DIFF
--- a/src/pya/unit_tests/pyaTests.cc
+++ b/src/pya/unit_tests/pyaTests.cc
@@ -106,10 +106,10 @@ PYTHONTEST (dbLayoutToNetlist, "dbLayoutToNetlist.py")
 PYTHONTEST (dbLayoutVsSchematic, "dbLayoutVsSchematic.py")
 PYTHONTEST (dbNetlistCrossReference, "dbNetlistCrossReference.py")
 PYTHONTEST (layLayers, "layLayers.py")
+PYTHONTEST (layPixelBuffer, "layPixelBuffer.py")
 PYTHONTEST (tlTest, "tlTest.py")
 #if defined(HAVE_QT) && defined(HAVE_QTBINDINGS)
 PYTHONTEST (qtbinding, "qtbinding.py")
 #endif
 
 #endif
-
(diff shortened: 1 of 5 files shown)


klayoutmatthias:
Hi Thomas,

here is the equivalent test from "layPixelBuffer.rb". It uses a synthetic image and does not suffer from changes in the drawing engine (which I assume may also render slightly different pixels on different machines). It should be good enough to test libpng integration:

```
    pb = RBA::PixelBuffer::new(10, 20)
    pb.transparent = true
    pb.fill(0xf0010203)

    ...

    png = nil
    begin
      png = pb.to_png_data
    rescue => ex
      # TODO: No PNG support -> error out
    end

    if png

      assert_equal(png.size > 20 && png.size < 200, true)  # some range because implementations may differ
      pb_copy = RBA::PixelBuffer.from_png_data(png)
      assert_equal(compare(pb, pb_copy), true)

      tmp = File::join($ut_testtmp, "tmp.png")
      pb.write_png(tmp)
      pb_copy = RBA::PixelBuffer.read_png(tmp)
      assert_equal(compare(pb, pb_copy), true)

    end
```

Best regards,

Matthias

thomaslima:
Hello! Sorry I didn't see this. The only issue, I think, is that ruby tests are not launched in the GitHub Actions for building python wheels. Would you like me to rewrite the test to match ruby's?

klayoutmatthias:
No worries :)

If possible, I'd advise to use the same scheme (in Python of course) - my personal experiences with stored layout images are pretty discouraging. The idea is to have a synthetic image for test and as PNG is lossless, the resulting image after the round trip should be identical to the original one.

Thanks,

Matthias

thomaslima:
@klayoutmatthias Done. I also fixed an issue with pya_tests.py. It was not loading all the tests (just the last one).

Best,
Thomas

PR #1347, opened 2023-04-23
update wheels to build on macOs M1

### LINKED ISSUE #1346: no wheels for M1
How can we make `pip install klayout` work on M1?

@thomaslima

joamatab:
fixes #1346

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       max-parallel: 12
       matrix:
-        os: [ubuntu-20.04, windows-2019, macos-11]
+        os: [ubuntu-20.04, macos-11]
     steps:
       - name: Free Disk Space (Ubuntu)
         if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       max-parallel: 12
       matrix:
-        os: [ubuntu-20.04, macos-11]
+        os: [ubuntu-20.04]
     steps:
       - name: Free Disk Space (Ubuntu)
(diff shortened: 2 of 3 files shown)


klayoutmatthias:
Could you elaborate what is the basic improvement that makes it work with M1? Or more specifically, why the previous version was not building arm64 wheels? Isn't there a minimum fix? And what do the unrelated entries mean, such as 'tool.cibuildwheel.overrides' and why was musllinux added and now it gets removed?

Honestly, I am a bit tired of this CI magic. It needs maintenance because they constantly change things there. It's taking too much of my limited time which I'd rather like to spend on features. Debugging is a pain and adding M1 in general means more support effort. 

Matthias

joamatab:
Hi Matthias,

we are at the python conference sprinting on gdsfactory and we realised that MacOs M1 users can't `pip install klayout`

we just  released new gdstk wheels for arm based Macos so that `pip install gdsfactory` works now
https://github.com/heitzmann/gdstk/actions/runs/4789080539

This PR would build wheels for klayout in M1

you can see arm wheels [here](
https://github.com/KLayout/klayout/actions/runs/4789102928/jobs/8516569537)

klayoutmatthias:
Maybe I'm blind, but I see only x86 wheels being built. I checked "Build wheels" and "Run actions/upload-artifact@v2".

joamatab:
great catch, just fix it, there will be uploaded it after this [commit](https://github.com/KLayout/klayout/pull/1347/commits/c180773546bd46046836c0f25573e7fb926b4be5)

https://github.com/KLayout/klayout/actions/runs/4791439354

klayoutmatthias:
Thanks, I'll check and upload manually. 

Regarding the last commit: wouldn't that be the only change needed for building arm64 wheels on MacOS? Plus maybe upgrade to pypa/cibuildwheel@v2.12.3?

joamatab:
yes, we just needed to add the flags for the macos-arm build

let me know when you upload them

thomaslima:
@klayoutmatthias I agree. I think there are a lot of unnecessary changes. I would appreciate a minimal change as well. I don't see the reason for a new `build_mac` recipe or removing the musllinux. Sounds like Joaquin was just using this PR to trigger multiple GitHub actions for testing purposes.

thomaslima:
Closed by #1378 and #1381 .
@joamatab Thanks for your help!

PR #1378, opened 2023-05-22
building arm64 wheels with cibuildwheel (uses intel runner)

thomaslima:
Currently cibuildwheel will attempt to build arm64 in an intel machine. Though this might work, I am not sure the external libraries would load correctly. I want to try and compare wheels built on an M1 machine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,12 +49,13 @@ jobs:
           HOST_CCACHE_DIR="$(ccache -k cache_dir)"
           mkdir -p $HOST_CCACHE_DIR
       - name: Build wheels  # check https://cibuildwheel.readthedocs.io/en/stable/setup/#github-actions
-        uses: pypa/cibuildwheel@v2.12.0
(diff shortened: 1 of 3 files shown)


klayoutmatthias:
Problem is that the tests won't run and there is no confirmation that the builds will work. Last time we tried that this wasn't the case.

thomaslima:
Hi Matthias.

I have completed the patch for side loading libpng in M1 wheels. The procedure is documented in the libpng-patch.sh file. I'm using home-brew's pre-compiled "bottle" to supply the binary library to be linked into the wheel. I tested GitHub compiled wheel for python 3.10 on my Mac. I'm reasonably confident the other wheels should work too. It's ready to merge.

I understand the risk of not testing for Apple Silicon specifically. But this should be better than the current situation anyway - as installing from source for the user also ignores testing. 

This patch is temporary until https://github.com/github/roadmap/issues/528 is closed. 

Wheels should be available in https://github.com/KLayout/klayout/actions/runs/5052077888

Closes #1346 #1347

PR #1391, opened 2023-05-29
autogenerate updated stubs when wheels in github

### LINKED ISSUE #1390: `scripts/make_stubs.sh` is not run automatically when releasing new pypi versions
I think when a new pypi package is published, the intention would be to update/create the python stub files. As far as I can tell, it would make sense to run these when/before running the wheel build(s). It seems thought that currently these files are tracked in the repository and don't get automatically updated.

Would it make sense to automate this? I am happy to make a PR if anyone can point me to where to do it. I tested it locally and I only could get it to run if the GUI has been built previously. But I am sure there is a better way.

@thomaslima 

(I noticed this when mypy was complaining that new API points aren't available, when they are there in the official docs and the local docs built into KLayout itself)

thomaslima:
Closes #1390

This will add a step between building wheels and publishing in pypi. In this step, it installs the built python package, regenerates the pyi stubs, and diff-checks with the current stubs in the repository. If there is a discrepancy, it pushes the changes via git to the same branch. Once the changes are pushed, the workflow can be rerun via GitHub if necessary without further commits.

Side effects:
- It pushes a new commit to the same branch, including the master branch if the action is set to run there.
- It pushes a commit to the PR and fails the test. The user needs to remember to pull from the PR branch. This will fail for PRs coming from forks because we wouldn't have permission to push to the base repository.
- I am not sure how this will behave in case the stubs are outdated and a release commit is pushed.  It will probably fail and not push the wheels to pypi (on GitHub actions).


thomaslima:
@klayoutmatthias How do we deal with the azure task? If you submit a release tag, this will fail but the azure will succeed and upload wheels to pypi it looks like. Maybe we can port the klayout build from azure to here?

The point is to prevent wheels and tarballs from being published if the stubs are outdated. With this PR, this will never happen if releases are created right after a PR merge (I usually do this on my other repos). But there could be a corner case where some manual commit to master would cause stubs to be outdated.

thomaslima:
How to re-run the action job after the pyi stub fails. :)
image

klayoutmatthias:
Hi Thomas,

thanks for your efforts - however I can hardly image that this might work. I think that is a kind of chicken-and-egg problem, isn't it?

I mean the autogeneration should happen before the release tag is made. Otherwise the new stubs will not be included in the tag. Or the tag has to be moved. Which creates other issues because on a release a lot more happens on my side, like building of signed Windows installer binaries etc. 

I admit I did not pay much attention to the stubs, but it's not a big deal to include the stub generation into my release checklist. That's a small step among others such as preparing release notes or signing the Windows binaries. 

What do you think?

Matthias

thomaslima:
Hello Matthias,

I understand the pain. I vote to modify the PR so the pyi stubs check doesn't fail. I have a few observations:

- If you're always careful about manually making sure the stubs are up-to-date, this workflow would have no side effects.
- If somehow you fall into the edge case of committing something to master that would cause new stubs, and not waiting till the workflow pushes new pyi stubs to tag the release (this will run on master if #1392 is accepted), then the wheels will be uploaded with outdated pyi stubs, and the issue will fix itself in the next release. This is an improvement on the current situation, where the wheels would still be incorrect and the issue undetected.

LG
Thomas

thomaslima:
I am closing this PR based on the comment in #1390. I will submit a separate PR with minor improvements on the make_stubs documentation. I am able to run it without having to do a full compilation of the klayout+pymod.

PR #1392, opened 2023-05-29
Update build.yml - enable builds on main branch

thomaslima:
pull requests should be able to access cache from main branch (caching is useless otherwise).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,8 +4,11 @@ name: Build Python Wheels
 
 on:
   pull_request:
-  # Running on PRs is enough
-  # push:
+  # Running on PRs and main branch only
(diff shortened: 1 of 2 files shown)


klayoutmatthias:
Do you have an idea why the operation was cancelled on the GitHub runners? It wasn't me, so maybe it was you?

continuous-integration/jenkins/pr-head is my own one - it's OK that was cancelled.

klayoutmatthias:
I'd merge is there is no reason to hold this PR back.

thomaslima:
Sorry. I cancelled manually because it would have been a waste of resources.

klayoutmatthias:
Very good. I'm in favor of not waisting resources. I'll merge then.

PR #1394, opened 2023-05-31
make_stubs now automatically detects installed klayout in python environment

thomaslima:
Solves a minor issue introduced in #1390. Now I can run make_stubs without calling it from the built pymod folder -- suffices to have klayout in a python site packages.


klayoutmatthias:
Very good. Thanks!

I'll wait if the CI complains and merge then.

PR #1524, opened 2023-11-06
Added the missing includes

### LINKED ISSUE #1523: Missing Includes
While building with Kubuntu 23.10, running the command
```
./build.sh -noruby -without-qt
```
causes a build error to occur. The error states missing definitions for <optional> and uint8_t. 

QMake suggests adding the following includes:
- GsiTypes.h
  - \<optional\>
  - \<cstdint\>
- GsiSerialization.h
  - \<optional\> 

After applying these changes to the header files, the compile completes successfully and the build begins. I suggest adding the includes to the header files as required.

wulfhednar02:
Added the missing includes <optional> and <cstdint> to GsiTypes.h.
Added the missing include <optional> to GsiSerialzation.h

fixes #1523

### DIFF
--- a/src/gsi/gsi/gsiSerialisation.h
+++ b/src/gsi/gsi/gsiSerialisation.h
@@ -31,6 +31,7 @@
 #include <list>
 #include <memory>
 #include <cstring>
+#include <optional>
 
 namespace gsi
 {
--- a/src/gsi/gsi/gsiTypes.h
+++ b/src/gsi/gsi/gsiTypes.h
@@ -33,6 +33,8 @@
 #include <vector>
 #include <list>
 #include <set>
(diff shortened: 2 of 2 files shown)


klayoutmatthias:
I have added `#include <cstdint>` but not the `<optional>` as this would require C++17. I think that is bug in a specific Qt version - a cannot reproduce the problem with the versions I am using - which is 6.6.1, 6.4.2 and 6.2.1.

I also had to fix some other things with "without-qt", like including DRC and LVS frameworks. So eventually the whole thing is bigger than this PR.

Thanks for providing it though. However, I'll close this as the relevant changes are in master now.

Best regards,

Matthias

PR #1575, opened 2023-12-14
Added script to build/deploy in MSYS2/ucrt64 environments

wsteffe:


wsteffe:
I am closing it because I want revert back last commit (added installer ...).
The installer is uploaded to the separate repo https://github.com/wsteffe/klayout_releases_ucrt64

wsteffe:
I am reopening tje pull request after haveìing reverted last commit (it may not be cancelled).
Anyway installer file was never pushed because large files can't be pushed on a public fork.

wsteffe:
Hi Matthias,
 to build and deploy in ucrt64 you have to execute script/deploy-win-ucrt64.sh from the shell ucrt64 (not msys shell as written at https://www.klayout.de/build.html for deploy-win-mingw.sh). 

Beside changing mingw_inst=/mingw64 to ucrt_inst=/ucrt64 and other similar changes I have also fixed a problem in finding Binary dependencies:

> libs=$(objdump -p $new_libs | grep "DLL Name:" | sort -u | sed 's/.*DLL Name: *//')

gave an error: argument list too long for objdump and the set of copied dlls was not complete.

PR #1577, opened 2023-12-18
Fixed wrong message about NOT found dll

wsteffe:
There was a typo in my previous commit. 
Mssages "NOT found xxx.dll"  were written because looking in the wrong folder.


klayoutmatthias:
Please review the file in master. I have changed a few things, i.e. generalization. It now works on my system and produces an installer that appears to be working. I'll add that build to the CI.

Issue #1356, opened 2023-04-30
Instance Property form: Array Instance col/row would be more intuitive then the current row/col


stefanottili:
In the current Instance Property form, the Array Instances entries are presented as rows/colums.
Please reverse this order, col/row (x, y) would be more intuitive..

It's also the order in which gds aref col/row number and coordinates are defined.
Oasis repetitions are defined as N-column (x) by M-row (y) dimensions.

klayoutmatthias:
This is how the three places for arrays now look like:

![image](https://github.com/KLayout/klayout/assets/25720841/7826c81a-9056-46a6-af3f-0950b6fac4ca)

![image](https://github.com/KLayout/klayout/assets/25720841/433cd497-033c-4e9d-b460-362525fd21da)

![image](https://github.com/KLayout/klayout/assets/25720841/491ec410-c819-4756-934b-896dd9c00cb7)

stefanottili:
great, 'now' as in 'upcoming', right ?
I find the col/row (x/y dimension) much more intuitive then today's master:
<img width="976" alt="image" src="https://github.com/KLayout/klayout/assets/25827408/e322c070-a6c8-49aa-9349-2851aeaf6c8d">

klayoutmatthias:
Yes, "now" means "soon" :)
I want to finalize and push out 0.29. It has been in the oven for too long already :(
Matthias

PR #1611, opened 2024-02-04
Fix tabbing order in "Editor Options / Instance / Array Instance" UI.

anntzer:
Currently, in the Array Instance sub-UI that appears when inserting a new array instance, trying to use the "tab" key to move through the entry widgets results in going through them in a strange order (column step x, column step y, row step y, row step x, dimension columns, dimension rows).  This patch declares the correct (natural) tabbing order for the widgets.

(However, I didn't actually check that this works because I haven't actually looked into the build scripts for klayout yet.)

### DIFF
--- a/src/edt/edt/EditorOptionsInst.ui
+++ b/src/edt/edt/EditorOptionsInst.ui
@@ -548,6 +548,23 @@
    <header>layWidgets.h</header>
   </customwidget>
  </customwidgets>
+ <tabstops>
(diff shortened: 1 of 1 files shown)


stefanottili:
it would be nice if this could be combined with https://github.com/KLayout/klayout/issues/1356

klayoutmatthias:
Cool :) Thanks!

PR #1617, opened 2024-02-08
allow build klayout for debian

joamatab:
Update instructions for debian users

@sebastian-goeldi


sebastian-goeldi:
Can we hold off on this if it's even desired ;). I have never tested this with a fresh install, whether these dependencies are actually all that's required. This is only my lazy self-interested patching to get latest klayout on my debian stable/testing

PR #1641, opened 2024-03-06
Klayout PyCell integration

ThomasZecha:
-added tl::optional as derivate of std::optional for c++17 and above, reduced
 implementation otherwise
-fixed missing include for c++17 and above
-added range constraints for PCell parameter

### DIFF
--- a/src/db/db/dbPCellDeclaration.h
+++ b/src/db/db/dbPCellDeclaration.h
@@ -30,6 +30,7 @@
 #include "dbLayout.h"
 #include "tlVariant.h"
 #include "tlObject.h"
+#include "tlOptional.h"
 
 namespace db
 {
@@ -69,7 +70,7 @@ public:
    *  @brief The default constructor
    */
   PCellParameterDeclaration ()
-    : m_hidden (false), m_readonly (false), m_type (t_none)
+    : m_hidden (false), m_readonly (false), m_type (t_none), m_range()
   {
     // .. nothing yet ..
   }
@@ -78,7 +79,7 @@ public:
    *  @brief The constructor with a name
    */
   PCellParameterDeclaration (const std::string &name)
-    : m_hidden (false), m_readonly (false), m_type (t_none), m_name (name)
+    : m_hidden (false), m_readonly (false), m_type (t_none), m_name (name), m_range()
   {
     // .. nothing yet ..
   }
@@ -87,7 +88,7 @@ public:
    *  @brief The constructor with a name, type and description
    */
   PCellParameterDeclaration (const std::string &name, type t, const std::string &description)
-    : m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description)
+    : m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description), m_range()
   {
     // .. nothing yet ..
   }
@@ -96,7 +97,7 @@ public:
    *  @brief The constructor with a name, type, description and default value
    */
   PCellParameterDeclaration (const std::string &name, type t, const std::string &description, const tl::Variant &def)
-    : m_default (def), m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description)
+    : m_default (def), m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description), m_range()
   {
     // .. nothing yet ..
   }
@@ -105,7 +106,7 @@ public:
    *  @brief The constructor with a name, type, description, default value and unit
    */
   PCellParameterDeclaration (const std::string &name, type t, const std::string &description, const tl::Variant &def, const std::string &unit)
-    : m_default (def), m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description), m_unit (unit)
+    : m_default (def), m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description), m_unit (unit), m_range()
   {
     // .. nothing yet ..
   }
@@ -114,7 +115,7 @@ public:
    *  @brief The constructor with a name, type and description and choice values / choice descriptions
    */
   PCellParameterDeclaration (const std::string &name, type t, const std::string &description, const std::vector<tl::Variant> &choices, const std::vector<std::string> &choice_descriptions)
-    : m_choices (choices), m_choice_descriptions (choice_descriptions), m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description)
+    : m_choices (choices), m_choice_descriptions (choice_descriptions), m_hidden (false), m_readonly (false), m_type (t), m_name (name), m_description (description), m_range()
   {
     // .. nothing yet ..
   }
@@ -248,6 +249,53 @@ public:
   }
 
   /**
+   *  @brief A enum describing the action of range violated
+   */
+  enum Action {
+    t_Reject = 1,   //  reject the parameter
+    t_Accept,       //  accept the parameter
+    t_Use_Default   //  use a default parameter (currently not supported)
+  };
+
+  void set_range (const tl::Variant& low, const tl::Variant& high, const tl::Variant& resolution, Action action = t_Reject)
+  {
+    m_range = Range(low, high, resolution, action);
+  }
+
+  typedef struct _Range {
+    _Range() :
+      m_low(),
+      m_high(),
+      m_resolution(),
+      m_action(t_Reject) {}
+
+    _Range(const tl::Variant& low, const tl::Variant& high, const tl::Variant& resolution, Action action = t_Reject) :
+      m_low(low),
+      m_high(high),
+      m_resolution(resolution),
+      m_action(action) {}
+
+    bool operator== (const _Range& other) const
+    {
+      return
+        m_low == other.m_low &&
+        m_high == other.m_high &&
+        m_resolution == other.m_resolution &&
+        m_action == other.m_action;
+    }
+
+    tl::Variant m_low;
+    tl::Variant m_high;
+    tl::Variant m_resolution;
+    Action m_action;
+  } Range;
+
+  const tl::optional<Range>& get_range() const
+  {
+    return m_range;
+  }
+
+   /**
    *  @brief Getter for the choice descriptions
    *
    *  The choice descriptions correspond to choice values. The descriptions
@@ -280,7 +328,8 @@ public:
            m_type == d.m_type &&
            m_name == d.m_name &&
            m_description == d.m_description &&
-           m_unit == d.m_unit;
+           m_unit == d.m_unit &&
+           m_range == d.m_range;
   }
 
 private:
@@ -291,6 +340,7 @@ private:
   type m_type;
   std::string m_name;
   std::string m_description, m_unit;
+  tl::optional<Range> m_range;
 };
 
 /**
--- a/src/db/db/gsiDeclDbLibrary.cc
+++ b/src/db/db/gsiDeclDbLibrary.cc
@@ -29,6 +29,7 @@
 #include "dbPCellDeclaration.h"
 #include "dbLibrary.h"
 #include "dbLibraryManager.h"
+#include "tlLog.h"
 
 namespace gsi
 {
@@ -719,12 +720,41 @@ void clear_choices (db::PCellParameterDeclaration *pd)
 
 void add_choice (db::PCellParameterDeclaration *pd, const std::string &d, const tl::Variant &v)
 {
-  std::vector<tl::Variant> vv = pd->get_choices ();
-  std::vector<std::string> dd = pd->get_choice_descriptions ();
-  vv.push_back (v);
-  dd.push_back (d);
-  pd->set_choice_descriptions (dd);
-  pd->set_choices (vv);
+  if (!pd->get_range().has_value())
+  {
+    std::vector<tl::Variant> vv = pd->get_choices ();
+    std::vector<std::string> dd = pd->get_choice_descriptions ();
+    vv.push_back (v);
+    dd.push_back (d);
+    pd->set_choice_descriptions (dd);
+    pd->set_choices (vv);
+  }
+  else
+  {
+    tl::warn
+        << "PCell parameter '" << pd->get_name() << "' has a range constraint, could not add choice '"
+        << v.to_string() << "'";
+  }
+}
+
+void set_range (db::PCellParameterDeclaration *pd, const tl::Variant& low, const tl::Variant& high, const tl::Variant& resolution, unsigned int action)
+{
+  if (pd->get_choices().empty() && pd->get_choice_descriptions().empty())
+  {
+    if ((pd->get_type() == db::PCellParameterDeclaration::t_int
+          || pd->get_type() == db::PCellParameterDeclaration::t_double) &&
+        !low.is_nil() && low.can_convert_to_double() &&
+        !high.is_nil() && high.can_convert_to_double())
+    {
+      pd->set_range(low, high, resolution, db::PCellParameterDeclaration::Action(action));
+    }
+  }
+  else
+  {
+    tl::warn
+        << "PCell parameter '" << pd->get_name() << "' has a choice constraint, could not add range '["
+        << low.to_string() << ", " << high.to_string() << "]'";
+  }
 }
 
 static unsigned int pd_type_int ()
@@ -770,6 +800,21 @@ static unsigned int pd_type_callback ()
 static unsigned int pd_type_none ()
 {
   return (unsigned int) db::PCellParameterDeclaration::t_none;
+}
+
+static unsigned int pd_action_reject ()
+{
+  return (unsigned int) db::PCellParameterDeclaration::t_Reject;
+}
+
+static unsigned int pd_action_accept ()
+{
+  return (unsigned int) db::PCellParameterDeclaration::t_Accept;
+}
+
+static unsigned int pd_action_use_default ()
+{
+  return (unsigned int) db::PCellParameterDeclaration::t_Use_Default;
 }
 
 db::PCellParameterDeclaration *ctor_pcell_parameter (const std::string &name, unsigned int type, const std::string &description)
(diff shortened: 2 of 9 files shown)


src/edt/edt/edtPCellParametersPage.cc
```diff
@@ -786,6 +804,11 @@ PCellParametersPage::get_parameters_internal (db::ParameterStates &states, bool
               ps.set_value (tl::Variant (v));
               lay::indicate_error (le, (tl::Exception *) 0);
 
+              if (p->get_range().has_value())
```
klayoutmatthias, line 807: Checking the value here does not have the effect of rejecting the wrong value as "ps.set_value" already happened.

src/edt/edt/edtPCellParametersPage.cc
```diff
@@ -762,6 +775,11 @@ PCellParametersPage::get_parameters_internal (db::ParameterStates &states, bool
...
```
klayoutmatthias, line 778: Checking the value here does not have the effect of rejecting the wrong value as "ps.set_value" already happened.

src/gsi/gsi/gsiTypes.h
```diff
@@ -35,6 +35,7 @@
 #include <set>
 #include <stdexcept>
 #include <cstdint>
+#include <optional>
```
klayoutmatthias, line 38: doesn't that need to be guarded against older C++?

src/pymod/distutils_src/klayout/dbcore.pyi
klayoutmatthias, line 1: This file is generated. Please do not edit. It will be generated before the next release.

src/tl/tl/tlOptional.cc
```diff
@@ -0,0 +1,32 @@
...
+
+*/
+
+
+#include "tlOptional.h"
+
+namespace tl
+{
+
+#if __cplusplus < 201703L
```
klayoutmatthias, line 28: Why not simply using the poor man's implementation always?

src/tl/tl/tlOptional.h
```diff
@@ -0,0 +1,160 @@
...
+        return &value();
+    }
+
+    const T* operator-> () const
+    {
+        return &value();
+    }
+
+private:
+    T a_value;
```
klayoutmatthias, line 109: Style: private members are called "m_..."

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -897,7 +955,10 @@ Class<db::PCellParameterDeclaration> decl_PCellParameterDeclaration ("db", "PCel
   gsi::method ("TypeLayer", &pd_type_layer, "@brief Type code: a layer (a \\LayerInfo object)") +
   gsi::method ("TypeShape", &pd_type_shape, "@brief Type code: a guiding shape (Box, Edge, Point, Polygon or Path)") +
   gsi::method ("TypeCallback", &pd_type_callback, "@brief Type code: a button triggering a callback\n\nThis code has been introduced in version 0.28.") +
-  gsi::method ("TypeNone", &pd_type_none, "@brief Type code: unspecific type")
+  gsi::method ("TypeNone", &pd_type_none, "@brief Type code: unspecific type") +
+  gsi::method ("REJECT", &pd_action_reject, "@brief Action type: reject violating parameter") +
```
klayoutmatthias, line 959: In other places, constants are not all upper case, so it would be more lik "Reject", "Accept" ... but anyway, I don't understand why we should have anything else than "Reject": so either a value is limited or not.

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -719,12 +720,41 @@ void clear_choices (db::PCellParameterDeclaration *pd)
 
 void add_choice (db::PCellParameterDeclaration *pd, const std::string &d, const tl::Variant &v)
 {
-  std::vector<tl::Variant> vv = pd->get_choices ();
-  std::vector<std::string> dd = pd->get_choice_descriptions ();
-  vv.push_back (v);
-  dd.push_back (d);
-  pd->set_choice_descriptions (dd);
-  pd->set_choices (vv);
+  if (!pd->get_range().has_value())
```
klayoutmatthias, line 723: I disagree with the way modifications are rejected. This basically will allow you to change a declaration from range+no choice to choices+no range only by first setting the range to "no range" and then adding choices. I prefer not to restrict setters and getters and rather give a warning in "check_values" if choices are present

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -874,6 +919,19 @@ Class<db::PCellParameterDeclaration> decl_PCellParameterDeclaration ("db", "PCel
     "This method will add the given value with the given description to the list of\n"
     "choices. If choices are defined, KLayout will show a drop-down box instead of an\n"
     "entry field in the parameter user interface.\n"
+    "If a range is already set for this parameter the choice will not be added and a warning message is showed.\n"
```
klayoutmatthias, line 922: Please do not implement it that way (see above)

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -874,6 +919,19 @@ Class<db::PCellParameterDeclaration> decl_PCellParameterDeclaration ("db", "PCel
...
+  ) +
+  gsi::method_ext ("set_range", &set_range, gsi::arg ("low"), gsi::arg ("high"), gsi::arg ("resolution"), gsi::arg ("action"),
```
klayoutmatthias, line 924: There is no corresponding getter.

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -874,6 +919,19 @@ Class<db::PCellParameterDeclaration> decl_PCellParameterDeclaration ("db", "PCel
...
+    "@brief Set a range constraint\n"
+    "This method will set a range constraint to the parameter with 'low' and 'high' as minimum and maximum value.\n"
+    "This range constraint will only be set if the parameter-, the low- and the high-type are numeric.\n"
+    "If a choice is already set for this parameter the range will not be set and a warning message is showed.\n"
```
klayoutmatthias, line 928: Do not implement that this way.

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -874,6 +919,19 @@ Class<db::PCellParameterDeclaration> decl_PCellParameterDeclaration ("db", "PCel
...
+    "The optional parameter 'resolution' will give a desired resolution value (currently not used).\n"
+    "The optional parameter 'action' determines the action to be invoked.\n"
+    "This action can be one of three values: REJECT, ACCEPT, USE_DEFAULT. If this failure action\n"
+    "parameter is not specified, then it will be REJECT by default.\n"
```
klayoutmatthias, line 932: There is no default given.

src/db/db/gsiDeclDbLibrary.cc
```diff
@@ -874,6 +919,19 @@ Class<db::PCellParameterDeclaration> decl_PCellParameterDeclaration ("db", "PCel
...
+    "If a range constraint is violated this parameter is marked wrong with violation hint in the\n"
```
klayoutmatthias, line 933: Language: "If the entered value is not within the specified range, a warning will be shown in the user interface".

src/db/db/dbPCellDeclaration.h
```diff
@@ -30,6 +30,7 @@
 #include "dbLayout.h"
 #include "tlVariant.h"
 #include "tlObject.h"
+#include "tlOptional.h"
```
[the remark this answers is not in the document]
ThomasZecha: That's a general style question: Use a specific value of item('s) to decide whether it is used or not, or wrap it in an optional? I prefer using optional because of it's speaking name and readability increasing concept: I'm optional.

src/db/db/dbPCellDeclaration.h
```diff
@@ -69,7 +70,7 @@ class DB_PUBLIC PCellParameterDeclaration
    *  @brief The default constructor
    */
   PCellParameterDeclaration ()
-    : m_hidden (false), m_readonly (false), m_type (t_none)
+    : m_hidden (false), m_readonly (false), m_type (t_none), m_range()
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/db/db/dbPCellDeclaration.h
```diff
@@ -248,6 +249,53 @@ class DB_PUBLIC PCellParameterDeclaration
   }
 
   /**
+   *  @brief A enum describing the action of range violated
+   */
+  enum Action {
+    t_Reject = 1,   //  reject the parameter
+    t_Accept,       //  accept the parameter
+    t_Use_Default   //  use a default parameter (currently not supported)
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/db/db/dbPCellDeclaration.h
```diff
@@ -248,6 +249,53 @@ class DB_PUBLIC PCellParameterDeclaration
   }
 
   /**
+   *  @brief A enum describing the action of range violated
+   */
+  enum Action {
+    t_Reject = 1,   //  reject the parameter
+    t_Accept,       //  accept the parameter
```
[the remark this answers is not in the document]
ThomasZecha: Coming from PyCell-API, e.g. for test-reason during development.

src/db/db/dbPCellDeclaration.h
```diff
@@ -248,6 +249,53 @@ class DB_PUBLIC PCellParameterDeclaration
...
+  void set_range (const tl::Variant& low, const tl::Variant& high, const tl::Variant& resolution, Action action = t_Reject)
+  {
+    m_range = Range(low, high, resolution, action);
+  }
+
+  typedef struct _Range {
+    _Range() :
+      m_low(),
+      m_high(),
+      m_resolution(),
```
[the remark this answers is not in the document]
ThomasZecha: Will be removed.

src/db/db/dbPCellDeclaration.h
```diff
@@ -248,6 +249,53 @@ class DB_PUBLIC PCellParameterDeclaration
...
+    t_Accept,       //  accept the parameter
+    t_Use_Default   //  use a default parameter (currently not supported)
+  };
+
+  void set_range (const tl::Variant& low, const tl::Variant& high, const tl::Variant& resolution, Action action = t_Reject)
+  {
+    m_range = Range(low, high, resolution, action);
+  }
+
+  typedef struct _Range {
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/db/db/dbPCellDeclaration.h
```diff
@@ -248,6 +249,53 @@ class DB_PUBLIC PCellParameterDeclaration
...
+    _Range() :
+      m_low(),
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/db/db/gsiDeclDbLibrary.cc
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/edt/edt/edtPCellParametersPage.cc
```diff
@@ -1085,6 +1108,23 @@ PCellParametersPage::states_from_parameters (db::ParameterStates &states, const
...
 }
 
+void
+PCellParametersPage::check_range(const tl::Variant& value, const db::PCellParameterDeclaration::Range& range)
+{
+  if (db::PCellParameterDeclaration::Action(range.m_action) == db::PCellParameterDeclaration::t_Reject)
+  {
+    if (!range.m_low.is_nil() && value < range.m_low)
+    {
+      throw tl::Exception(tl::to_string (tr("Range violation: value < low")));
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/edt/edt/edtPCellParametersPage.cc
```diff
@@ -1085,6 +1108,23 @@ PCellParametersPage::states_from_parameters (db::ParameterStates &states, const
...
+    }
+
+    if (!range.m_high.is_nil() && range.m_high < value)
+    {
+      throw tl::Exception(tl::to_string (tr("Range violation: value > high")));
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/edt/edt/edtPCellParametersPage.cc
```diff
@@ -1085,6 +1108,23 @@ PCellParametersPage::states_from_parameters (db::ParameterStates &states, const
   }
 }
 
+void
```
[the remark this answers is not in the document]
ThomasZecha: Will be adjusted.

src/edt/edt/edtPCellParametersPage.cc
ThomasZecha: The gui will reject in that case. But will move to above ps.set_value...

src/edt/edt/edtPCellParametersPage.cc
ThomasZecha: The gui will reject in that case. But will move to above ps.set_value...

src/gsi/gsi/gsiTypes.h
ThomasZecha: Yes, will be adjusted.

src/pymod/distutils_src/klayout/dbcore.pyi
ThomasZecha: Sorry, but that could not by deduced by a human reader. A first line comment 'This file is generated...' may help.
Will be adjusted.

src/tl/tl/tlOptional.cc
ThomasZecha: Counter question: Why using std::string and not a poor man implementation of it. :o)
And the next inclined reader will ask: Why isn't std::optional used when available? :o)

Now serious: I prefer not to re-invent something if it's already there.

src/tl/tl/tlOptional.h
ThomasZecha: Will be adjusted.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Will be adjusted.
Accept coming from PyCell-API, e.g. for test-reason during development.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Will be adjusted.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Will be adjusted.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Will be added.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Will be adjusted.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Coming from PyCell API. Will be removed.

src/db/db/gsiDeclDbLibrary.cc
ThomasZecha: Will be adjusted.

ThomasZecha:
Thanks for the critical review.

I see a range constraint on the same conceptual level as choice constraint:

1. limiting the possible values of a parameter
2. are intuitively recognizable as such by the user
3. are difficult for the user to operate incorrectly (Provide instant feedback if)

So why implement a range differently to a choice?
Of course a range could be implemented by coerce_parameters but in a much less fashion regarding 2. and 3.

By the way: coerce_parameters delivers only a list of the parameter values, but not corresponding names. This could make an implementation quite unwieldy. Why not using a map?

klayoutmatthias:
The choices are actually meant for mode parameters (strings), not for numerical values. Also, I see that many constraints are more complex, like a maximum area, which limits the product of two parameters. Such constraints are pretty common and cannot be captured by a simple range.

"coerce_parameters" also plays a different role: as the name implies it provides some normalization and some warranties to the layout production code. "coerce_parameters" is also called under other circumstances, specifically right before layout production happens. So even if the parameters are change by script, the validity of the parameter set is established before "produce" is called.

If you rely on the limits implied by the range, it is not enough to capture the edit case. However, if you see the ranges as visual aids only, capturing them during edits is good enough, but not warranty is implied. This is important to note for the authors of the production function so their code can be aware of that case and raise an exception for example. In that context, getters for the range limit are very important, as they would allow friendly PCell instantiation code to check the limits before setting a parameter value.

Side note about the return value of "coerce_parameters": parameters are always lists internally, corresponding to the declaration order. This is for the internal interfaces and I cannot change that without breaking low-level API compatibility. For PCell authors there is the "PCellDeclarationHelper" wrapper which provides a more intuitive access layer.

There are some further comments I should make:
* Please do not add features that are "not supported yet". Unless you want to become permanent maintainer of the code yourself, everyone coming along such features will expect me to eventually implement it.
* The range feature should be noted in the documentation too. The main documentation for PCell authors is embedded into `src/db/db/built-in-macros/pcell_declaration_helper.lym` and `src/db/db/built-in-pymacros/pcell_declaration_helper.lym`.
* When you test your implementation, please try the following scenario: create a layout with a PCell and certain parameters and no range set. Save this layout. Then change the PCell implementation such that you introduce a range and the saved PCell violates the range. When you load the layout again, the implementation is supposed to indicate the current parameters as "outside the limits" when you check the PCell parameters in the UI. Unless you cancel the dialog, the implementation should not allow you to commit the original values. Specifically, this means that range checks are not only required on edits, but also on initialization of the UI. Please check, if your implementation does so.
* Another related topic are the stored "recent" parameters of PCells. KLayout offers to use recently used PCell configurations on the "Recent" tab in the editor options. In the above scenario, it may happen that a PCell parameter set becomes invalid after changing the range and restarting the application. The application should behave reasonable in that case and refuse placing such a PCell.

I am mentioning the "change of limits" case as I came across that problem a number of times in Cadence world. This is a real issue and not a hypothetical one.

I also need to say - and maybe it becomes clearer now - that although the feature may look trivial, it is not. The current code is the result of many iterations with users and constraint management was one of the key issues.

I am sorry for the many topics and modification requests, but I could have told you that in advance if I had a chance to.

Matthias

PR #1695, opened 2024-04-28
Trying to fix MacOS builds of Python modules

klayoutmatthias:


klayoutmatthias:
Found solution, removing PR

PR #1711, opened 2024-05-18
Replace "removed" Ruby method "file.exists?" with "file.exist?"

BWsix:
This PR replaces the only `file.exists?` method in the project with `file.exist?`.

`file.exists?` was deprecated [and removed from Ruby version 3.2.0](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/). I encountered this error when trying to open a DRC script with Ruby version 3.3.2 installed.


klayoutmatthias:
Thanks for that PR. I did not notice `File.exists?` is deprecated.

Matthias

Issue #1747, opened 2024-06-18
CplxTrans (and VCplxTrans?) on negative floats/ints


sebastian-goeldi:
Hello Matthias,

I don't know how much work it would be, but would it be possible to add or change the CplxTrans, and I think the VCplxTrans would also be necessary, to allow the usage on signed int instead of only unsigned int?

Reasoning being I added this https://github.com/gdsfactory/kfactory/issues/404 and didn't read the API properly before :P (the function there just takes any of the `__mul__` overloaded types and returns `CplxTrans(layout.dbu) * input` (or inverted in the `to_um` case))

I can work around it, but it would be nicer if the transformations directly support it. Of course only if the change is reasonable and wouldn't jeopardize other API.

PR #1753, opened 2024-06-22
Fixing issue #1747 (signed coordinate transformations with Trans/DTra…

klayoutmatthias:
…ns/CplxTrans etc.)


sebastian-goeldi:
Thank you Matthias. I was prepared to just have this handled in python. This is great!

Issue #1756, opened 2024-06-25
Tools --> Manage Packages Timeout


Kazzz-S:
The original issue: https://www.klayout.de/forum/discussion/2525/

The `Tools --> Manage Packages` operation often failed with a timeout (10s) in a particular Internet environment.
More details follow after [this link](https://www.klayout.de/forum/discussion/comment/10982/#Comment_10982).

Kazzz-S:
Enhance by #1764

Kazzz-S:
This enhancement seems missing in the ChangeLog for 0.29.3 2c0a7f79

klayoutmatthias:
No, I wrote the changelog too early :)

I will add it. Thanks for the hint!

Matthias

PR #1762, opened 2024-06-29
Introduce a new environment variable "KLAYOUT_INPUT_HTTP_STREAM_PRIVA…

Kazzz-S:
…TE_DATA_TIMEOUT" to control the timeout for XML document fetching

This is my idea to address the issue #1756.

### DIFF
--- a/src/tl/tl/tlHttpStreamQt.cc
+++ b/src/tl/tl/tlHttpStreamQt.cc
@@ -28,6 +28,9 @@
 #include "tlObject.h"
 #include "tlTimer.h"
 #include "tlSleep.h"
+#include "tlEnv.h"
+#include <sstream>
+#include <iomanip>
 
 #include <QNetworkAccessManager>
 #include <QNetworkReply>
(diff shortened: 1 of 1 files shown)


klayoutmatthias:
HI @Kazzz-S,

thanks for this PR.

It is fine with me to call the variable simply "KLAYOUT_HTTP_TIMEOUT". That is more generic and fits the purpose.

Matthias

klayoutmatthias:
Hi @Kazzz-S,

I hope it's not a problem I created an alternative proposal here: https://github.com/KLayout/klayout/pull/1764

Reason is that there is also the non-Qt, curl-based HTTP(S) implementation and I wanted both to behave the same. The curl version was not thrown an exception in case of timeouts, so I added this too. I also added a test and named the variable "KLAYOUT_HTTP_TIMEOUT".

Maybe you can take a look there.

Thanks,

Matthias

Kazzz-S:
Hi @klayoutmatthias,

Thank you for taking care of the non-Qt case, too.
Of course, the new implementation is OK with me.

Regards,
Kazzz-S

Kazzz-S:
Hi @klayoutmatthias,

I have tested #1764 on macOS, which works fine.

Thank you.
Kazzz-S

PR #1778, opened 2024-07-06
trying to fix MacOS builds: delocate-wheel was erroring out as libpng…

klayoutmatthias:
… was sometimes on the current MacOS version


klayoutmatthias:
The problem was maybe twofold:
* On MacOS 13 libpng was installed as version for MacOS 13
* cibuildwheels is using delocate 0.11 for Python >3.6 which checks for compatibility
* Also, cibuildwheels configures MacOS 10.9 as detault version, so delocate errored out
* Disabling the MacOS version configuration only made delocate pick MacOS 13 as the minimum requirement
* The resulting build cannot be tested with cibuildwheels as it demands MacOS 10.9 compatibility
* Solution is to disable MacOS version configuration for delocate and rename the resulting wheel, so it is recognized as MacOS 10.9 version.
Clean solution was to find a way to install libpng for MacOS 10.9 for the MacOS 13 builds.

PR #1785, opened 2024-07-09
Fixed DeepShapeStore.subcircuit_hierarchy_for_nets getter

lanserge:
Previously introduced DeepShapeStore.subcircuit_hierarchy_for_nets getter is actually a copy&past of the setter.
As result the DeepShapeStore class has two `subcircuit_hierarchy_for_nets` setters which results in `ERROR: RuntimeError: Ambiguous overload variants - multiple setter declarations match arguments in DeepShapeStore.subcircuit_hierarchy_for_nets` when setter called from scripts. This error makes `subcircuit_hierarchy_for_nets` setter unusable/unavailable from scripts.

The fix updates `DeepShapeStore.subcircuit_hierarchy_for_nets` getter declaration, which makes setter and getter work correctly.

### DIFF
--- a/src/db/db/gsiDeclDbDeepShapeStore.cc
+++ b/src/db/db/gsiDeclDbDeepShapeStore.cc
@@ -190,7 +190,7 @@ Class<db::DeepShapeStore> decl_dbDeepShapeStore ("db", "DeepShapeStore",
     "\n"
     "This attribute has been introduced in version 0.28.4"
   ) +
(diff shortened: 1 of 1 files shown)


klayoutmatthias:
Thanks for this PR. This was an actual bug.

I just wonder: this is a very specialized feature. It affects the way the DRC "nets" method returns the net shapes in deep mode. I think the default value of "false" is the only thing that currently makes sense. I have used "true" in only one case, which is a unit test.

Matthias

Issue #1793, opened 2024-07-19
inconsistent results in width_check when calling by keyword argument (python)


tvt173:
Hi @klayoutmatthias ,

I found a troublesome bug today when using a keyword argument to set the `min_projection` in a `width_check` for a `Region` object in python. I find that I get consistent results when calling using a positional argument but inconsistent results when using a keyword argument. Do you know why this might be?

Please see the example script below:

```python
import klayout.db as kdb
import numpy as np

def always(value_dbu):
    # min_projection is not set
    return region.width_check(value_dbu)

def never(value_dbu):
    # setting min_projection via a positional argument
    return region.width_check(value_dbu, False, kdb.Metrics.Euclidian, np.deg2rad(90), value_dbu)

def sometimes(value_dbu):
    # setting min_projection via a named argument
    return region.width_check(value_dbu, min_projection=value_dbu)

filename = r"troublesome_taper.oas"
l = kdb.Layout()
value_dbu = 1000
l.read(filename)
layer = l.layer(8, 0)
region = kdb.Region(l.begin_shapes(l.top_cell(), layer=layer))
for func in [always, never, sometimes]:
    results = []
    for i in range(30):
        result = len(func(value_dbu))
        results.append(str(result))
    print(f"{func.__name__}: " + "".join(results))
```

Expected Output:
```console
always: 111111111111111111111111111111
never: 000000000000000000000000000000
sometimes: 000000000000000000000000000000
```

Actual Output:
```console
always: 111111111111111111111111111111
never: 000000000000000000000000000000
sometimes: 011000100001001100000100000000
```

I'm using the standalone klayout python package v0.29.2 on Windows 11. 

Sample gds: [troublesome_taper.zip](https://github.com/user-attachments/files/16302459/troublesome_taper.zip)

sebastian-goeldi:
Your `never` case might be flawed:

The ignore angle should be in degrees not radian:

From the docs:

```
"ignore_angle" specifies the angle limit of two edges. If two edges form an angle equal or above the given value, they will not contribute in the check. Setting this value to 90 (the default) will exclude edges with an angle of 90 degree or more from the check. Use nil for this value to select the default.
```

So for proper comparison it should be `90` or even better `None`

Does this change the output of `never`?

tvt173:
nice catch @sebastian-goeldi ! indeed it changes the results for the "never" case, but it does not change the fact that "sometimes" behaves inconsistently

updated script:
```python
import klayout.db as kdb
[...]

PR #1795, opened 2024-07-19
Fixing issue #1793 (problem with default arguments in Python)

klayoutmatthias:


tvt173 (approved):
thanks for the fix and additional tests! the changes look reasonable as far as i can tell. let me know if you'd like me to test anything on my end

Issue #1836, opened 2024-08-27
Allow the usage of `Cell.shapes(LayerInfo(...))`


sebastian-goeldi:
Hi Matthias,

I'll cut to the chase directly: Would it be possible to allow `Cell.shapes` to take a LayerInfo object and have it resolve it to the layer_index internally? If easily possible this would also be nice for `Cell.begin_shapes_rec(...)` and its friends. But those are less affected than shapes due to infrequent use of those.

_The longer story:_

I am currently working on expanding the use of (gds)pdks (well, pdk is a very advanced word for where we are in photonics, but you get the general idea)  across multiple layouts (in order to essentially  separate cells into more or less standard cells and cells generated specifically for the tapeout such as new structures for tests etc).

One of the big struggles I encountered is that most of my end-users do not really understand why  a layer index is kind of bound to the layout. So, I have essentially rewritten (or rather hidden deeper in the code)  an enum for the layer indexes to mainly use LayerInfo objects instead.

But now I have the problem of having to explain to users why they have to do `c.shapes(c.layout().find_layer(LayerInfo(...)))` (of course this is a bit less explicit than here, but for illustration purposes it's in pure klayout code).

Best,
Sebastian

klayoutmatthias:
Hi @sebastian-goeldi,

That is possible of course, but there are many places where layer indexes are used and some don't have a layout.

If you need support for explaining this decision - here is the motivation: There are anonymous layers. Whenever you do some computations, you may encounter cases when you need temporary layers. DRC decks use plenty of them for example. A temporary layer does not have coordinates like layer or datatype number. So LayerInfo is not a good way of identifying them. The layer index is more generic and allows addressing anonymous layers too.

Matthias

sebastian-goeldi:
Hi Matthias,

Yes, I am fully aware that there are cases.

I am merely requesting this as I think it would be generally a good usability improvement for pure layout work (as in constructing the layout), and less so for any checking/verification work.

It would be nice to have the layoutinfo options as second class citizens alongside the index, purely for the cell functions (and maybe layout functions) where applicable.

But if you think it's not really feasible or out of scope, that's totally fine. To me (from a very surface perspective view), it seems like it wouldn't be a gigantic change, but I could totally be wrong. Feel free to close the issue in that case.

klayoutmatthias:
[...]

PR #1849, opened 2024-09-07
Implemented solution for issue #1836 (Allow the usage of Cell.shapes(…

klayoutmatthias:
…LayerInfo(...)))

- For Cell#shapes, LayerInfo is an accepted argument now
- If the layer does not exist, it is created (non-const version) or an error is raised (const version)
- Cell#clear also accepts a LayerInfo object
- Layout#clear and Layout#delete_layer also accept a LayerInfo parameter

Rationale for supporting the clear and delete_layer methods: These cases do nothing if the layer does not exist, so there is a benefit of using them: it is not required to check first if the layer exists.

### DIFF
--- a/src/db/db/gsiDeclDbLayout.cc
+++ b/src/db/db/gsiDeclDbLayout.cc
@@ -1103,6 +1103,30 @@ void break_polygons1 (db::Layout *layout, size_t max_vertex_count, double max_ar
   db::break_polygons (*layout, max_vertex_count, max_area_ratio);
 }
 
+void delete_layer_from_info (db::Layout *layout, const db::LayerProperties &info)
+{
+  int li = layout->get_layer_maybe (info);
+  if (li >= 0) {
+    layout->delete_layer ((unsigned int) li);
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/db/db/gsiDeclDbLayout.cc
```diff
@@ -1963,18 +1987,38 @@ Class<db::Layout> decl_Layout ("db", "Layout",
...
-    "@param layer_index The index of the layer to delete.\n"
+    "@param layer_index The index of the layer to clear.\n"
   ) +                               
+  gsi::method_ext ("clear_layer", &clear_layer_from_info, gsi::arg ("layer"),
+    "@brief Clears a layer\n"
+    "\n"
+    "This variant takes a \\LayerInfo object to address the layer.\n"
+    "It does nothing if no layer with these attributes exists.\n"
+    "\n"
+    "This variant was introduced in version 0.26.7.\n"
```
jukkarabina, line 1998: The version number should be 0.29.7, right @klayoutmatthias ?

PR #1916, opened 2024-11-01
Some add-ons for potential application in the PEX context

klayoutmatthias:


martinjankoehler:
Hi @klayoutmatthias,

there seems to be an issue now with the `space_check` (for sidewall),
it works if I only have 1 layer, but if I have a second layer all the markers are empty.

Example GDS:
[near_body_shield_li1_m1.gds.gz](https://github.com/user-attachments/files/17757340/near_body_shield_li1_m1.gds.gz)

Bildschirmfoto 2024-11-14 um 21 20 34

klayoutmatthias:
That's easy ...

67/20 is covering all of 68/20, so there is no space :)

Matthias

martinjankoehler:
> That's easy ...
> 
> 67/20 is covering all of 68/20, so there is no space :)
> 
> Matthias

Right, I need to call it for each layer separately (earlier we used multiple layers (acting as nets), but only non-overlapping ones).

Issue #1930, opened 2024-11-18
Selection of structures results in thick black lines


hpretl:
When I use KLayout 0.29.8 inside a Docker VM based on Ubuntu 22.04LTS, and when I am on a **Arm-based Mac** (_not happening on Intel-based Macs_) then I get a thick black selection frame when clicking on a layout structure:

![image](https://github.com/user-attachments/assets/a6c13b76-76a2-4b10-842c-b35ec00c902b)

Interestingly, when I hover the mouse over a structure, there is also a black line indicating a to-be selected layout structure, but this black “frame” looks OK:

![image](https://github.com/user-attachments/assets/fd6943ff-b4c5-46bc-b05c-15a8bc5b8ca6)

@klayoutmatthias Any idea what is going on? The problem is that this thick black border on selected items is really annoying, as it overlays other things that would be necessary to see. All other graphical elements in KLayout look fine, just this selection border is too wide.

As a side note, turning off “with halo” or playing around with line width in the Setup Selection does not help.

martinjankoehler:
FYI, I can also reproduce it using VNC (instead of X11):
![Bildschirmfoto 2024-11-18 um 20 20 11](https://github.com/user-attachments/assets/d56b497c-d512-49c6-a9fd-1882f9e61692)

martinjankoehler:
Compiler warning hints at the reason:
```
../../../src/laybasic/laybasic/layMarker.cc: In member function \u2018void lay::MarkerBase::get_bitmaps(const lay::Viewport&, lay::ViewObjectCanvas&, lay::CanvasPlane*&, lay::CanvasPlane*&, lay::CanvasPlane*&, lay::CanvasPlane*&)\u2019:
../../../src/laybasic/laybasic/layMarker.cc:326:33: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  326 |   int line_width = m_line_width < 0 ? mp_view->default_marker_line_width () : m_line_width;
      |                    ~~~~~~~~~~~~~^~~
../../../src/laybasic/laybasic/layMarker.cc:327:35: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  327 |   int vertex_size = m_vertex_size < 0 ? mp_view->default_marker_vertex_size () : m_vertex_size;
      |                     ~~~~~~~~~~~~~~^~~
../../../src/laybasic/laybasic/layMarker.cc:328:22: warning: comparison is always false due to limited range of data type [-Wtype-limits]
[...]

PR #1932, opened 2024-11-19
Bugfix/Issue 1930: Fixed type from char to int

martinjankoehler:
Fixes issue #1930


klayoutmatthias:
Ah, the famous "char is signed or unsigned" problem :)

I thought that has been settled as clang and gcc treat it the signed on both Linux and Windows, but ARM appears to be an exception.

Usually I try to avoid that, but here it escaped me. "signed char" would have solved the issue too, but now we're prepared for line widths >=128. Lovely :)

Thanks for the patch.

Best regards,

Matthias

Issue #1963, opened 2024-12-30
Add lock to Cell


sebastian-goeldi:
Hi Matthias,

Would it be possible to add a "lock" to the Cell class?

This might sound like a weird request, so let me explain the background a bit first. In gdsfactory/kfactory when we create cells, we usually do that with cached functions. Now, sometimes we want to modify cells in a certain way, e.g. add some cladding or exclude layers. This often happens with helper functions. When the caching has been done, it is (almost always) a bad idea to further alter the cell. Therefore, kfactory implemented a "locked" state on its implementation of the Cell, KCell. This lock prevents some stuff that is accessible to the user such as creation of ports (essentially pins with a width and angle) or transform (there is a slight modification to the function in order to transform to transform the ports as well). What of course doesn't (easily) work is to lock the addition of shapes, as those are completely managed by klayout. Ideally, we can enhance this lock to cover most bases (as always, there will be ways to get to the gun and shoot your foot, just want to avoid the non-obvious cases).

I understand if this would go too far into the core of `db` and would require lots of engineering or hog too much resources. But I thought it's still worth asking ;)

P.S. I am not sure what it is related to, but recently I have seen some weird behavior of the layout changes state. I couldn't determine yet where exactly it's coming from, but I am getting some of the "layout state is not writable - forcing update" (sorry forgot exact error message) on write. I think it either comes from the TilingProcessor or from dangling references to shape/instance iterators is my guess (I usually only see this message with slightly faulty code or in failing tests). Will the latter cause the layout to be in the state equivalent to `Layout.start_changes()`? If not , did any of the TilingProcessor code change recently? I haven't seen anything in the patch notes. (Sorry, was not sure where to add this, so I added it here, but I can make a new issue of course if desired, or forum post).

klayoutmatthias:
Hi @sebastian-goeldi,

A general write lock is possible I think, but whether that is possible on cell level I have some doubts. For shapes I think so, but for instances, the situation is more complicated. Instances not only change the cell per se, but also the relationship between cells and KLayout does not differentiate between cell and the cell graph. So locking the cell graph may be possible, but locking one cell against adding or removing instances which another cell isn't locked, probably is not possible.

[...]

PR #1965, opened 2025-01-04
Feature/issue 1963

klayoutmatthias:
Implemented a solution for #1963 (locking cells)
    
A cell can be locked using
```
cell.locked = true
```
and unlocked again using
```
cell.locked = false
```
Also, `cell.is_locked?` can be used to test the locked state.
    
In locked state writing shapes and instances is forbidden
and doing so would raise an exception.
    
Also, cells cannot be deleted when locked. However, Layout#clear
and Layout#_destroy are always available.
    
Cells can still be renamed, even if locked.


sebastian-goeldi:
Thank you so much Matthias! This is awesome

joamatab:
Thank you Matthias, this would be very helpful for GDSFactory, looking forward to the next KLayout release!

PR #2003, opened 2025-03-22
Timer::memory_size() implementation for macOS

martinjankoehler:
## Test

`Makefile`:
```Makefile
all: waster mem

waster: waster.cpp
	g++ -o waster waster.cpp

mem: mem.cpp
	g++ -o mem mem.cpp
```

`waster.cpp`:
```c++
#include <iostream>
#include <cstdint>

int main(int argc, char **argv) {
    const size_t buffer_size_mb = 200;
    const size_t buffer_size = buffer_size_mb * 1024 * 1024;
    uint8_t *buffer = new uint8_t[buffer_size];
    for (size_t i=0; i < buffer_size; ++i) { buffer[i] = i & 0xFF; }

    std::cout << "allocated " << buffer_size_mb << " MB, press enter to terminate …" << std::endl;
    std::string s; std::cin >> s;

    delete[] buffer;
    return 0;
}
```

Bildschirmfoto 2025-03-22 um 23 55 54

```c++
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/types.h>
#include <libproc.h>

void get_memory_usage(pid_t pid) {
    struct proc_taskinfo taskinfo;
    int result = proc_pidinfo(pid, PROC_PIDTASKINFO, 0, &taskinfo, sizeof(taskinfo));

    if (result <= 0) {
        perror("proc_pidinfo failed");
        return;
    }

    printf("PID: %d\n", pid);
    printf("Resident Set Size (RSS): %.2f MB\n", taskinfo.pti_resident_size / (1024.0 * 1024.0));
    printf("Virtual Memory Size (VMS): %.2f MB\n", taskinfo.pti_virtual_size / (1024.0 * 1024.0));
}

int main(int argc, char *argv[]) {
    if (argc != 2) {
        fprintf(stderr, "Usage: %s <pid>\n", argv[0]);
        return 1;
    }

    pid_t pid = atoi(argv[1]);
    get_memory_usage(pid);
    return 0;
}
```

Bildschirmfoto 2025-03-22 um 23 56 59

### DIFF
--- a/src/tl/tl/tlTimer.cc
+++ b/src/tl/tl/tlTimer.cc
@@ -33,6 +33,7 @@
 #  include <mach/mach.h>
 #  include <sys/times.h>
 #  include <unistd.h>
+#  include <libproc.h>
 #else
 #  include <sys/times.h>
 #  include <unistd.h>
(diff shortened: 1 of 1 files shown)


martinjankoehler:
OK, in the Macro Development of a running instance it looks like this:
Bildschirmfoto 2025-03-23 um 00 10 54

klayoutmatthias:
Thanks! But I guess I should change Linux to rss instead of vsize to be consistent.

My understanding of "WorkingSetSize" in Windows is that it's actually close to rss.

Matthias

PR #2047, opened 2025-05-11
Add beta and free-threaded versions to build.yml

clin1234:


klayoutmatthias:
Hi @clin1234,

thanks for the PR, but I am not in favor of adding experimental things. This CD, not CI. It includes an upload step. If one of the betas fail, it will not upload.

I have already enough trouble keeping the builds alive. I don't want to add complexity here.

After all, I am author, maintainer and supporter in one person, and that in my spare time.

Matthias

PR #2052, opened 2025-05-18
 Adding freecad python modules

wsteffe:
The change of PYTHONPATH made in script deploy-win-ucrt64 is required for importing
 FreeCAD, Part, and other FreeCAD submodules.

FreeCAD must be installed in the MSYS2 before running that script.This is done with command:  "pacman -S mingw-w64-ucrt-x86_64-freecad".

I have also done a full upgrade of my MSYS2 system with "pacman -Suy" because it was very outdated. Then I have built klayout 30.1 and It seems working well but there is a problem: FreCAD python modules were included in subfolders Ext and Mod of bin-release-win64-ucrt but not included in the NSIS package.

### DIFF
--- a/scripts/deploy-win-mingw.sh
+++ b/scripts/deploy-win-mingw.sh
@@ -176,6 +176,15 @@ if ! [ -e $target/klayout.exe ]; then
 fi
 
 # ----------------------------------------------------------
+# FreeCAD
+if [ "$MSYSTEM" == "UCRT64" ]; then
+cp /ucrt64/bin/FreeCAD.exe  $target
+cp /ucrt64/bin/FreeCAD.pyd  $target
+cp /ucrt64/bin/FreeCADCmd.exe $target
+cp /ucrt64/bin/FreeCADGui.pyd $target
+fi
+
+# ----------------------------------------------------------
 # cert.pem
 
 echo "Installing cert.pem .."
--- a/scripts/deploy-win-mingw.sh
+++ b/scripts/deploy-win-mingw.sh
@@ -175,14 +175,6 @@ if ! [ -e $target/klayout.exe ]; then
   exit 1
 fi
 
-# ----------------------------------------------------------
-# FreeCAD
-if [ "$MSYSTEM" == "UCRT64" ]; then
-cp /ucrt64/bin/FreeCAD.exe  $target
-cp /ucrt64/bin/FreeCAD.pyd  $target
-cp /ucrt64/bin/FreeCADCmd.exe $target
-cp /ucrt64/bin/FreeCADGui.pyd $target
-fi
 
 # ----------------------------------------------------------
 # cert.pem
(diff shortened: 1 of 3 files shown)


klayoutmatthias:
Hi @wsteffe,

Thanks for the PR, but you're saying that the FreeCAD modules is non-standard in a way that it needs additional subfolders?

If true, that's unfortunate, as the whole process of collecting the dependencies is based on a traversal of the link dependencies. If there are files that are not part of these dependencies they are not seen. So far that was not the case with all the other Python modules.

Matthias

wsteffe:
It is not standard but it seems to me that  the process of collecting the dependencies worked well anyway. In fact after deploy-win-ucrt64 have created the folders Ext and Mod in bin-release-win64-ucrt:
Screenshot from 2025-05-19 05-51-01
The main freecad module (import FreeCAD) is placed in Ext/freecad while all FreeCAD submodules are placed in Mod/subfolder (in example Mod/Part for import Part).
I think that it is should be sufficient to add Ext and Mod (with all their content) to the list of packaged folders (but I do not know where this list is defined).

wsteffe:
No sorry, I think you were right saying that there is a problem with the traversal of the link dependencies.
In fact only python files directly referenced in python import where discovered but their dependences (in example OpanCascade library that is used in FreeCAD) are missing. 
I am expecting that the complete dependency tree to bee a very large amount of data.
So probably it would make sense to put it in a different (addon/optional) package that my be named klayout_FreeCAD similarly to what I did in https://github.com/wsteffe/layoutDD_FreeCAD.

klayoutmatthias:
A special treatment just for FreeCAD, and that's only for UCRT64?

The dependency traversal should work recursively, but it does not work on extension modules that are loaded dynamically. I assume this is the case here. 

I am quite skeptical whether the solution you propose is sustainable. I assume the folder structure is internal to FreeCAD and they might change it. Honestly, that is not something I want to maintain continuously. The numerous variables within the various systems from GitHub to PyPI make deployment an adventure already. I am seriously looking for package maintainers that have bandwidth for these kind of specialties, but personally I am at the very edge of my resources.

Matthias

wsteffe:
Hi Matthias, I made a last commit in case you reconsider including FC modules in klayout ucrt distribution. It is only for ucrt because in a Linux system you may use the FreeCAD modules installed in the system (in example with ubuntu you get them with "sudo apt install freecad"). Mingw64 is not supported by FreeCAD maintainers and I think it is regarded as obsolete.

klayoutmatthias:
Hi @wsteffe,

Let me try things before going ahead. I will do a system upgrade and see how it goes. Python 3.12 shouldn't be an issue in general, unless they have already upgraded to 3.13 :(

Matthias

wsteffe:
Hi Matthias, after several experiments I have discovered that many files were not found in the dependency search for FreeCAD modules. The number of required files (see at https://github.com/wsteffe/klayout_FreeCAD) has grown a lot (up to 1.6 GB) so I am dismissing the request to include FreeCAD modules in klayout distribution.

klayoutmatthias:
Hi @wsteffe,

thanks for letting me know.

I have tried upgrading MSYS2 to upgrade Python. I've got 3.12.10 now, but 32bit Windows support is lacking in most parts - specifically Python modules.

The best approximation I can make as of now is to continue doing 32bit builds with Python 3.11.

Matthias

PR #2053, opened 2025-05-19
Fix logic error in hex2int.

hzeller:
This was probably never noticed as it was the last branch and would behave benign on valid input.


klayoutmatthias:
Ah, right. Stupid mistake.

Thanks for fixing that!

Matthias

PR #2056, opened 2025-05-20
Allow no-GIL wheels to be uploaded to PyPI as well

clin1234:
Supercedes #2047 by not building wheels for pre-release CPython versions

As evident by https://github.com/clin1234/klayout/actions/runs/15031698862, free-threaded wheels for 3.13 builds successfully


klayoutmatthias:
I am sorry, I don't want to mess with my builds. There is not enough space on PyPI to accommodate much more versions, and I want my deployment to be stable. I don't have any bandwidth for experiments here.

PR #2084, opened 2025-07-05
Feature/issue 2044

klayoutmatthias:


stefanottili:
Thanks for working on this, oh so close ...

Please see that attached testcase, c contains top with an unresolved reference to test.
I would expect that reference to be resolved with the "merged" aka "blend-mode 0" a.oas+b.oas.

[test_merge_file.tgz](https://github.com/user-attachments/files/21081452/test_merge_file.tgz)

Issue #2108, opened 2025-08-06
Build failure on macOS Sequoia: Undefined symbol gsi::cls_decl<db::Net>() in dbMeasureEval.o


Kazzz-S:
Hello @klayoutmatthias,

First of all, thank you for your efforts in preparing the 0.30.3 release.

**Description of the issue**
Building the current `master` branch on macOS Sequoia (Intel) fails with an undefined symbol error related to `db::Net` when linking `libklayout_db`.

**Environment:**
- macOS Sequoia 15.6 (Intel)
- Apple clang version 17.0.0 (clang-1700.0.13.5)
- Target: x86_64-apple-darwin24.6.0
- KLayout commit: 10a34cca (master branch, 0.30.3-to-be)
  (built locally with a slight modification: ef365b55 to bypass the link error)
- Build system: `build4mac.py` with the default options.

**Error log:**
```
Undefined symbols for architecture x86_64:
"gsi::ClassBase const* gsi::cls_decldb::Net()", referenced from:
db::MeasureNetEval::net_func() const in dbMeasureEval.o
ld: symbol(s) not found for architecture x86_64
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

**Analysis (by ChatGPT):**
The function `MeasureNetEval::net_func()` (in `src/db/dbMeasureEval.cc`) calls:
```
return tl::Variant::make_variant_ref(n->second);
```
where n->second is a const db::Net*.
This requires a GSI binding for db::Net (gsi::cls_decl<db::Net>()).
I confirmed that such a binding does not exist in the current source tree by searching (`grep -R "cls_decl<db::Net>" src/`), which returned no results.

This causes a link error on macOS (and possibly other platforms with stricter linking).
It may have been macOS’s stricter linker that surfaced this issue.

**Workaround (by ChatGPT):**
As a temporary fix, replacing the line (Left) with a dummy `return tl::Variant();` (Right) allows the build to complete.
<img width="1402" height="433" alt="Image" src="https://github.com/user-attachments/assets/a64790f6-f657-4bcc-b087-05d40bec98af" />

**Rebuild**
Succeeded 😄 
<img width="1012" height="640" alt="Image" src="https://github.com/user-attachments/assets/6206faba-447f-4e09-8e5f-258e80bd4467" />

Best regards,
Kazzz-S

klayoutmatthias:
Hi @Kazzz-S,

[...]

PR #2109, opened 2025-08-06
Trying to fix issue #2108

klayoutmatthias:


stefanottili:
builds on M1 homebrew

PR #2189, opened 2025-10-18
Feature/issue 2180

klayoutmatthias:


stefanottili:
in viewer mode, one can select and query the unresolved references, and the number of selected references is reported in the status line, but they don't get "highlighted" (in my case) green as they are in edit mode.

klayout -ne unresolved_ref.oas

klayoutmatthias:
You're right. Thanks for noticing. The last commit should fix that issue.

Best regards,

Matthias

PR #2193, opened 2025-10-19
Feature/issue 2147

klayoutmatthias:
The solution implements a merge for auto-loaded library files. Layout files with the same LIBNAME are merged into a single library.

In this implementation, the first of those files (in lexical order) "wins": the database unit is taken from that file. Also further files are added in lexical order. Secondary files do not delete or overwrite cells already present - so again, earlier files have higher priority.


ayala-io:
Thank you so much for this @klayoutmatthias !!
Looking forward to the 0.30.5 release!

PR #2283, opened 2026-02-24
Avoid crashing when clicking reload

abdelq:
For context, this was happening to me on macOS with files being written to a temporary directory and no longer existing there. I would attempt to reload with the notification and KLayout would crash instead of giving an error message.


klayoutmatthias:
The reason why signal/slot exceptions are undefined is that Qt offers to move signals between threads or delay execution. Both cases don't handle exceptions, that's why. Single-threaded direct signal/slot calls just get routed over C++ code generated by MOC and unless they have changed that significantly in Qt 6.11, I can't see why that should lead to lost exceptions. 

The change you propose is okay and I will merge it, but I still think the reason sits deeper. So I'd expect that you see more crashes in other cases not related to signals and slots.

Matthias

PR #2370, opened 2026-06-05
 CI/CD triggers to skip builds on non-source changes

nikosavola:
Relating to https://github.com/KLayout/klayout/issues/2368#issuecomment-4625100031

Updates CI/CD pipelines to prevent expensive and slow Python wheel builds when changes are limited to documentation, samples, or legacy config files.

#### Changes
 - GitHub Actions (.github/workflows/build.yml): Added a changes filter job using [dorny/paths-filter](https://github.com/dorny/paths-filter). The `build` and `make_sdist` jobs now run conditionally. This uses job-level checks to ensure skipped builds report as successful and satisfy branch protection rules.
 - Azure Pipelines (azure-pipelines.yml): Added native trigger and pr path exclusions for non-source folders (docs/*, samples/*, macbuild/*, etc.) and project metadata files to skip Windows builds on unrelated modifications.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,34 @@ on:
     types: [published]
 
 jobs:
+  changes:
+    name: Check for source changes
+    runs-on: ubuntu-latest
+    outputs:
+      code: ${{ steps.filter.outputs.code == 'true' || github.event_name == 'release' }}
+    steps:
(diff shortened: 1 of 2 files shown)


klayoutmatthias:
Sounds good. Thanks! Let's give it a try.

PR #2371, opened 2026-06-05
Use native concurrency groups in GitHub CI build workflow

nikosavola:
Replaced the obsolete [`styfle/cancel-workflow-action`](https://github.com/styfle/cancel-workflow-action) step with native GitHub Actions concurrency configuration in the wheel-building workflow. This is recommended in the action's README page as well.


klayoutmatthias:
Also sounds good. Let's also try ...

PR #2378, opened 2026-06-23
Add clang-format config tuned to KLayout C++ style

nikosavola:
In this PR, we add a .clang-format config tuned to the existing C++ style in KLayout. Then, we also add a CI check that fails if the the codebase is not formatted. Finally, in a single commit we run clang format on all the source files. 

Let me know if the clang-format file needs any fine tuning.


klayoutmatthias:
Hi @nikosavola,

Although the changes are fairly conservative and the style is nicely matching, it's a tough decision to merge such a commit.

Main reason is that I have a number of old branches open which I want to merge when time comes, and I don't want to drown in conflicts.

Matthias

nikosavola:
> Hi @nikosavola,
> 
> Although the changes are fairly conservative and the style is nicely matching, it's a tough decision to merge such a commit.
> 
> Main reason is that I have a number of old branches open which I want to merge when time comes, and I don't want to drown in conflicts.
> 
> Matthias

Understandable. Supporting autoformatting would be nice to have. I guess it could either be introduced later or alternatively add the `.clang-format` config but not enforce it for everything yet in CI. Maybe format only files when they are changed?

PR #2395, opened 2026-07-18
Add drag-and-drop reordering of layout view tabs

nharjee:
## Summary

Enables drag-and-drop reordering of the layout view tabs in the main window. Previously the tabs could not be rearranged; this makes the tab bar movable and keeps the underlying view model in sync when tabs are dragged.

## Details

The tab index is used as a positional key into `mp_views` and the seven parallel widget stacks (view / layer-control / layer-toolbox / hierarchy / libraries / editor-options / bookmarks), so reordering the `QTabBar` alone would desync them.

- `MainWindow`: enable `QTabBar::setMovable(true)` and connect `QTabBar::tabMoved(int,int)` to a new `MainWindow::tab_moved(from, to)` slot that reorders `mp_views` and every widget stack to match the new tab order, then re-raises the current widgets.
  - Qt emits `tabMoved` before `currentChanged`, so any subsequent view selection operates on the already-reordered stacks.
  - The current view is tracked by object (`LayoutView::current()`), so it is never lost across a reorder — only its index changes.
- `ViewWidgetStack` and `ControlWidgetStack`: add a `move_widget(from, to)` helper that reorders the internal widget vector.

Net change is additive: 6 files, +64 lines.

## Testing

- Built on macOS (Qt 6, `build4mac.py`).
- Full unit-test suite (`ut_runner` via `macQAT.py`) run before and after: the same pre-existing set of failures (unrelated `rba`/`lvs`/`drc`/`tl` reference-data and environment-specific tests) appears with and without this change; no new failures are introduced.
- Manually verified: drag to reorder, then switch / clone / close tabs — the tab↔view mapping stays consistent.

## Known cosmetic note

On macOS, while a tab is actively being dragged, Qt's built-in movable-tab drag rendering tints the tabs flanking the lifted tab blue (the platform style's `dragInProgress` paint path, using a grabbed-pixmap overlay for the moving tab). This is Qt-native behavior triggered simply by enabling `setMovable(true)` — it occurs before any reorder logic runs, is purely visual, and reverts on drop. Left as-is.

### DIFF
--- a/src/lay/lay/layControlWidgetStack.cc
+++ b/src/lay/lay/layControlWidgetStack.cc
@@ -127,6 +127,17 @@ void ControlWidgetStack::remove_widget(size_t index)
   update_geometry ();
 }
 
+void ControlWidgetStack::move_widget (size_t from, size_t to)
+{
+  if (from == to || from >= m_widgets.size () || to >= m_widgets.size ()) {
+    return;
+  }
+
+  QWidget *w = m_widgets [from];
+  m_widgets.erase (m_widgets.begin () + from);
+  m_widgets.insert (m_widgets.begin () + to, w);
+}
+
 void ControlWidgetStack::raise_widget(size_t index)
 {
   mp_current_widget = 0;
--- a/src/lay/lay/layControlWidgetStack.h
+++ b/src/lay/lay/layControlWidgetStack.h
@@ -45,6 +45,7 @@ public:
 
   void add_widget (QWidget *w);
   void remove_widget (size_t index);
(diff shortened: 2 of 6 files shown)


klayoutmatthias:
The PR looks okay, but let me tell you that I am absolute not in favor of AI-authored PRs. I am basically inclined to reject them unconditionally, because I don't consider them trustworthy and I don't want to accept PRs that take me more time to review than it took to create them.

I think there is more to consider with respect to tab reordering. I will take a look, but with low priority.

Matthias

PR #2447, opened 2026-09-18
Add option to use layer names in DXF export (#2439)

### LINKED ISSUE #2439: layers name in DXF file export
Hi Matthias,

The DXF export is a very useful feature for assembly. We export a DXF file with the die size and the pads, eventually some alignment marks.
However, a small improvement would be nice to the DXF import.
Assume we export the layer : 18/0  which name is :  PAD

Then in the DXF the layer name will appear as :
<code>L18D0</code>
It would be nice if it could appear in the DXF file with its name :
<code>PAD</code>

Maybe, it can be an option in the DXF export : **<code> using the layers name </code>**

BRgds,
Laurent

goanscryca:
This PR addresses #2439 by adding support for exporting DXF layers with their original layout layer names instead of prefixing layer/datatype numbers (e.g. `PAD` instead of `L18D0_PAD`).

### Changes
1. **`DXFWriterOptions` (`dbDXFFormat.h`)**:
   - Added `use_layer_names` boolean flag (defaults to `false` for backwards compatibility).
2. **`DXFWriter::write` (`dbDXFWriter.cc`)**:
   - Uses `db::SaveLayoutOptions::LP_AssignNameWithPriority` when `use_layer_names` is true, preserving layer names directly as DXF layer names.
3. **GSI & Python/Ruby bindings (`gsiDeclDbDXF.cc`, `dbcore.pyi`, `laySaveLayoutOptions.rb`)**:
   - Added `dxf_use_layer_names` / `dxf_use_layer_names=` accessors on `SaveLayoutOptions`.
4. **GUI Option Page (`DXFWriterOptionPage.ui`, `layDXFWriterPlugin.cc`)**:
   - Added "Use layer names (don't prefix with layer/datatype numbers)" checkbox to the DXF export options dialog.
5. **XML persistence (`dbDXF.cc`)**:
   - Added XML serialization for `use-layer-names`.
6. **CLI buddy options (`bdWriterOptions.h`, `bdWriterOptions.cc`, `bdBasicTests.cc`)**:
   - Added `--use-layer-names` flag to generic writer options.
7. **Unit tests (`dbDXFWriterTests.cc`, `bdBasicTests.cc`, `laySaveLayoutOptions.rb`)**:
   - Added `TEST(LayerNames)` verifying DXF export with and without `use_layer_names`.

Fixes #2439.


Kazzz-S:
@goanscryca,

Thank you for implementing PR #2447!
I built and tested the PR on macOS, and my testing shows it provides a direct solution to the original request in #2439.

First, I confirmed the new DXF writer option:
issue-2439-DXFWriter-option

### A. Layers without names
When the original KLayout layers do not have layer names:
issue-2439-A-in

the exported DXF keeps the conventional `L<layer>D<datatype>` names:
For example:

- `1001/0` → `L1001D0`
- `2001/1` → `L2001D1`
- `3001/1` → `L3001D1`

issue-2439-A-out

----

### B. Layers with names
When the KLayout layers have names:
issue-2439-B-in

with **"Use layer names (don't prefix with layer/datatype numbers)"** enabled, those names are directly used as the DXF layer names:
For example:

- `AAA 1001/0` → `AAA`
- `BBB 2001/1` → `BBB`
- `CCC 3001/1` → `CCC`

issue-2439-B-out

----

This is exactly the behavior @laurentc2 was looking for in the original issue: existing KLayout layer names are now preserved directly in the DXF export, while unnamed layers retain their conventional layer/datatype-based names.

Thank you very much for implementing this feature!

Best regards,
Kazzz-S

